PR #85, opened 2021-08-11
cleanup namespace for hydra_zen.typing

rsokl:
Moved the implementations of our protocols to `hydra_zen.typing._implementations` so that the namespace of `hydra_zen.typing` consists only of the types that we choose to ship. 

The only compatibility-breaking change is that `hydra_zen.typing.DataClass` was made to be strictly internal, since I don't want to purport to be providing a satisfactory protocol for describing dataclasses in general.

I am on the fence of exposing `hydra_zen.typing.Partial`, for similar reasons to making `DataClass` internal, but it does seem useful to facilitate convenient annotations for our users like 

```python
def custom_instantiate(x: PartialBuilds[T]) -> Partial[T]:
    ...
```

Also, I am more confident that our `Partial` protocol is sufficiently general.

### DIFF
--- a/src/hydra_zen/experimental/_implementations.py
+++ b/src/hydra_zen/experimental/_implementations.py
@@ -13,8 +13,8 @@ from hydra.plugins.sweeper import Sweeper
 from hydra.types import HydraContext, RunMode
 from omegaconf import DictConfig, OmegaConf
 
-from .._hydra_overloads import instantiate
-from ..typing import DataClass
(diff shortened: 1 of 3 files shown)


jgbos:
I guess I don't have a strong feeling about what to expose.  I haven't thought about how I might try to extend to a custom instantiate but I like that idea, which means I would in favor of exposing `Partial`.

jgbos (approved):

PR #94, opened 2021-09-10
Adds `hydra_zen.get_target`

### LINKED ISSUE #92: New feature: `hydra_zen.get_target`
`hydra_zen.get_target` would enable users to access (and import) the target of a config.

```python
>>> Conf = builds(dict)
>>> get_target(Conf)
<class 'dict'>
```

This is useful for things like loading a lightning module from a checkpoint, via a hydra yaml.

```python
>>> ExpConfig = load_from_yaml("my_exp.yaml")
>>> LitModule = get_target(ExpConfig.lit_module)  # get lightning module class
>>> module_kwargs = instantiate(builds(dict, builds_bases=(ExpConfig.lit_module,))) 
>>> LitModule.load_from_checkpoint(checkpoint_path="./chk.ckpt, **module_kwargs)
```

Without `get_target`, one would need to do

```python
>>> LitModule = type(instantiate(ExpConfig.lit_module))
```
which is a bit wasteful, since you have to initialize the whole module just to get the type.

I can imaging that `get_target` could be handy in simpler cases as well.

rsokl:
Closes #92 

Adds `hydra_zen.get_target`:

```python
>>> from hydra_zen import builds, just, get_target
>>> get_target(builds(int))
int
>>> get_target(just(str))
str
```

First of all.. sorry for somehow touching 12 files in this PR that adds only one function. 😞 

It turned out that this was a bit more delicate than I had anticipated. The biggest complicating factor was distinguishing `Just` and `PartialBuilds` instances from `Builds`. In these cases, we don't want to merely retrieve `_target_`, since it points to helper functions - `get_obj` and `partial` respectively. So I had to take care to handle cases where a dataclass *looks* like, say, `Just` (i.e. it has `_target_` and `_path_` attributes), but is in fact not a `Just` (e.g. its `_target_` does not point to `get_obj`). Otherwise we could mess up on a `Builds` that simply has a `path` attribute.

Some additional changes that I made while working on this:
- Broadened the type-overloads on `instantiate` so that `omegaconf` containers are accomodated
- Improves pyright tests
- Fixed some inconsistencies in our `PartialBuilds` and `Just` protocols and added regression tests. 

Note: our overloads are getting to be quite unruly... we probably need to move to type stubs moving forward.


jgbos (approved):
Yeah lots of changes.  It all looks good reading through the changes.

PR #104, opened 2021-09-18
Deprecate `builds(target=<target>, ...)`; make target positional-only

### LINKED ISSUE #100: Change signature of builds to not occupy name "target"
Our signature for `builds`, `builds(target, ...)`, occupies the name `target` such that the following function is impossible to use with `builds`:

```python
def f(*, target): return target

builds(f, target=1)  # TypeError: builds() got multiple values for argument 'target'
builds(f, 1)  # TypeError: f takes 0 positional args, but 1 were specified via `builds`
```

I...definitely should have seen this coming. The ideal fix for this would be to make `target` be a positional-only argument, but we have to [wait until Python 3.8](https://www.python.org/dev/peps/pep-0570/) until we can do that.

In the meantime, we can opt to rename this in the fashion of the other names that are reserved by `builds`. I.e. call it `hydra_target`.

I have never encountered anyone using `builds(target=...)` explicitly, but I would still like to handle this as responsibly/gracefully as is reasonable. I'll try to come up with a plan for raising a deprecation warning for cases where people used `target` as a named argument.

rsokl:
Closes #100 

Before:

```python
builds(target=dict, a=1, b=2)  #  allowed

builds(dict, target=1)  # raises (multiple values for target)
```

After:

```python
builds(target=dict, a=1, b=2)  #  Deprecated. Will raise: missing required positional arg `hydra_target`

builds(dict, target=1)  # allowed

builds(hydra_target=dict)  # error missing required positional-only arg `hydra_target`
```


src/hydra_zen/structured_configs/_implementations.py
```diff
@@ -424,13 +451,13 @@ def builds(
 
     Returns a dataclass object that configures ``target`` with user-specified and auto-populated parameter values.
 
-    The resulting dataclass is specifically a structured config [1]_ that enables Hydra to initialize/call
+    The resulting dataclass is specifically a structured co nfig [1]_ that enables Hydra to initialize/call
```
rsokl, line 454:
```suggestion
    The resulting dataclass is specifically a structured config [1]_ that enables Hydra to initialize/call
```

jgbos (approved):

PR #110, opened 2021-09-29
Migrate all zen-processing to `_zen_xx`-named fields

rsokl:
## Migrating to `_zen_xx` fields
Per @Jasha10 's recommendation, we want to avoid potential future-conflicts with Hydra keeping away from the `_xx_` naming scheme. Instead we will exclusively reserve fields with the naming scheme `_zen_xx`.

Thus specifying `hydra_partial=True` will now occupy `_zen_target` and `_zen_partial`, instead of `_partial_target_` . Note that this PR preserves compatibility with configs/yamls that *do* specify `_partial_target_`. I.e. this is not a compatibility-breaking change 🥳 .

## Introducing common machinery for all zen-specific processing
This PR also introduces `hydra_zen.funcs.zen_processing`, which will be our "uber" processing function that handles all pre/post processing surrounding instantiation. Here, it simply implements `hydra_partial=True`, but in an upcoming PR, it will also facilitate `hydra_meta`. This will make it much easier to make multiple zen features (e.g. specifying `hydra_partial=True` *and* `hydra_meta=...`) compose, and intermingle.

## Variable-name squatting
Lastly, we will officially have `builds` reserve all names of the form `hydra_xx` so that we never have to worry about "taking" these names from users in future. We also 

## Summary

Before:

```python
>>> from hydra_zen import builds, to_yaml
>>> c = builds(dict, a=1, hydra_partial=True)
>>> print(to_yaml(c))
_target_: hydra_zen.funcs.partial
_partial_target_:
  _target_: hydra_zen.funcs.get_obj
  path: builtins.dict
a: 1
```

After:

```python
>>> print(to_yaml(c))
_zen_target: builtins.dict
_zen_partial: true
a: 1
```

Before:

```python
>>> builds(dict, hydra_something=1)  # OK
>>> builds(dict, _zen_partial=1)  # OK
```

After:

```python
>>> builds(dict, hydra_something=1)  # Error: we reserve all hydra_xx names
>>> builds(dict, _zen_partial=1)  # Error: we reserve all _zen_xx names
```

### DIFF
--- a/.github/workflows/tox_run.yml
+++ b/.github/workflows/tox_run.yml
@@ -170,4 +170,4 @@ jobs:
         python -m pip install --upgrade pip
         pip install -e .
     - name: Run pyright
-      run: pyright --lib tests/annotations/
+      run: pyright --lib tests/annotations/ src/
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,16 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# hydra-zen docs
+
+docs/source/_build
+docs/builds
+
+# pycharm
+.idea/
+
+# pyright
+node_modules/
+
+scratch/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,16 @@
+repos:
+-   repo: https://github.com/psf/black
+    rev: 21.8b0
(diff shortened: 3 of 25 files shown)


jgbos (approved):

rsokl:
🎉

PR #125, opened 2021-10-07
Use type-checking hack to make builds-sig legible in jupyter

rsokl:
This is [the second time](https://github.com/rsokl/MyGrad/blob/4256b6f286e52eaa819fd6657decc705ce3dd32a/src/mygrad/typing/_array_like.py#L51-L56) I have encountered this problem...

The type-annotations for `zen_wrapper` do a great job during static-checking, but they make an absolute mess of the signature of `builds`, when viewed from common editors:

### Pre-PR
**VSCode (Pylance)**
vscode

**PyCharm**
pycharm

**Jupyter**
jupyter

### Post-PR
The `TYPE_CHECKING` hack + identifying this [pylance behavior](https://github.com/microsoft/pylance-release/issues/1914#issuecomment-938187778) and updating the default value of `zen_wrappers` to `tuple()` leads to...

New Jupyter:

image

New VSCode:

image

(no impact on PyCharm)

### DIFF
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -42,22 +42,18 @@ except ImportError:  # pragma: no cover
 
 _T = TypeVar("_T")
 _T2 = TypeVar("_T2", bound=Callable)
-_Wrapper = Callable[[_T2], _T2]
-
+ZenWrapper = Union[
+    None,
+    Builds[Callable[[_T2], _T2]],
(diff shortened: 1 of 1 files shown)


rsokl:
@jgbos  this is ready for review

jgbos:
Sorry, I didn't see the notice till now.

jakebailey:
FWIW we'll probably have more improvements (in addition to the one I talked about) for next week's Pylance, which may make this easier without workarounds.

PR #122, opened 2021-10-06
Implements zen-wrappers

rsokl:
This adds a new feature: `builds( .... zen_wrappers=<..>)`.

This is an extremely powerful feature that enables one to inject modifications to the instantiation process by including wrappers in a target's configuration.

## The Basics

Whereas

```python
Conf = builds(f, *args, **kwargs)
instantiate(Conf)
```
calls `f(*args, **kwargs)`, the following config with wrappers

```python
Conf = builds(f, *args, **kwargs, zen_target=[w1, w2, w3])
```

will instantiate as:

```python
f = w1(f)
f = w2(f)
f = w3(f)
f(*args, **kwargs)
```

The wrapper paradigm enables arbitrary pre-processing, post-processing, and transformation of the target and its inputs/outputs.

### Seeing it in Action
Suppose we have a function that takes in a temperature in Farenheit, and converts it to Celcius.

```python
def faren_to_celsius(temp_f):
    return ((temp_f - 32) * 5) / 9
```

```python
>>> AsCelcius = builds(faren_to_celsius)
>>> instantiate(AsCelcius, temp_f=32)
0.0
```

But we discover that we need the temperature to be converted to Kelvin instead. We can modify our config to include the following wrapper:

```python
>>> def change_celcius_to_kelvin(celc_func):
...     def wraps(*args, **kwargs): 
...         return 273.15 + celc_func(*args, **kwargs) 
...     return wraps
```
```python
>>> AsKelvin = builds(faren_to_celsius, zen_wrappers=change_celcius_to_kelvin)
>>> instantiate(AsKelvin, temp_f=32)
273.15
```

### Another use case

In the next PR, users will be able to add pydantic validation to the instantiation process.

```python
from pydantic import PositiveInt
from hydra_zen.experimental.third_party.pydantic import validates_with_pydantic

def f(x: PositiveInt):
    return x

Conf = builds(f, zen_wrappers=validates_with_pydantic)

instantiate(Conf, x=10)  # ok
instantiate(Conf, x=-10)  # pydantic raises validation error
```

### Wrappers are configurable
The wrappers themselves can be made configurable:

```python
ConfWrapper = builds(some_wrapper, ....)

Conf = builds(f, zen_wrapper=ConfWrapper)
```

### `None` can be used as a placeholder for wrappers

`None` can be be interspersed among wrappers; it will be ignored. This means that `None` can be used as a placeholder for wrappers that are going to be toggled on/off from the CLI, via config groups. See the next subsection for an example.

[...]

PR #126, opened 2021-10-08
Implements `validates_with_pydantic`

rsokl:
[As promised](https://github.com/mit-ll-responsible-ai/hydra-zen/pull/122), this PR adds `validates_with_pydantic`, which is designed to be used with the new zen-wrappers feature. Everything that you instantiate can now be validated by [pydantic](https://pydantic-docs.helpmanual.io/)!

** Edit: moved `hydra_zen.experimental.third_party` to `hydra_zen.third_party` **

## The Basics

```python
>>> from hydra_zen.third_party.pydantic import validates_with_pydantic
>>> from pydantic import PositiveInt
>>> def f(x: PositiveInt): return x
>>> val_f = validates_with_pydantic(f)
>>> f(-100)
-100
>>> val_f(-100)
ValidationError: 1 validation error for F (...)
```

Using `validates_with_pydantic` with a class.

```python
>>> class A:
...     def __init__(self, x: PositiveInt): ...
>>> validates_with_pydantic(A)  # wrapping occurs in-place
__main__.A
>>> A(-10)
ValidationError: 1 validation error for Init
```

## Using `validates_with_pydantic` with hydra-zen

Obviously, this is designed to be used with the ``zen_wrappers`` feature of `builds`.

```python
>>> from hydra_zen import builds, instantiate
>>> # instantiations of `conf` will be validated by pydantic
>>> conf = builds(f, populate_full_signature=True, zen_wrappers=validates_with_pydantic)
>>> instantiate(conf, x=10)
10
>>> instantiate(conf, x=-2)
ValidationError: 1 validation error for F (...)
```

Note that pydantic's data-coercion is ensures smooth compatibility with Hydra.
I.e. lists will be coerced to the appropriate annotated sequence type.

```python
>>> def g(x: tuple): return x  # note the annotation
>>> validates_with_pydantic(g)([1, 2, 3])  # input: list, output: tuple
(1, 2, 3)
```

### pydantic-dependent configs
This introduces an optional dependency – leveraging this feature requires pydantic to be installed. It also means that yamls produced via `validates_with_pydantic` will also carry this dependency. But this is nice! You get validation even when you modify your yaml!

We could also design `validates_with_pydantic` to be a no-op if pydantic is not installed; this would enable folks to load yamls without having pydantic available. I am curious to get opinions on this. I like the idea that using `validates_with_pydantic` means that you *always* get validation.

### A known issue

The decorator `pydantic.validate_arguments` is what we utilize to obtain this validation. It raises a custom `ValidationError`, which has a weird interface. [Hydra's instantiate mechanism catches and re-raises all errors, so that it can modify the error message](https://github.com/facebookresearch/hydra/blob/85323ea76dc9b5f848e45b26d4d283fa90221901/hydra/_internal/instantiate/_instantiate2.py#L61-L67](https://github.com/facebookresearch/hydra/blob/85323ea76dc9b5f848e45b26d4d283fa90221901/hydra/_internal/instantiate/_instantiate2.py#L61-L67
). But `ValidationError` can't be called as `ValidationError(e)`, thus the attempt at re-raising causes a `TypeError`.

So the validation error actually looks like this:

```
ValidationError                           Traceback (most recent call last)
~\.conda\envs\hydra-zen\lib\site-packages\hydra\_internal\instantiate\_instantiate2.py in _call_target(_target_, *args, **kwargs)
     61 
---> 62         return _target_(*args, **kwargs)
     63     except Exception as e:

c:\users\ryan soklaski\hydra_utils\src\hydra_zen\funcs.py in zen_processing(***failed resolving arguments***)
    105         return _functools.partial(obj, *args, **kwargs)
--> 106     return obj(*args, **kwargs)

~\.conda\envs\hydra-zen\lib\site-packages\pydantic\decorator.cp38-win_amd64.pyd in pydantic.decorator.validate_arguments.validate.wrapper_function()

~\.conda\envs\hydra-zen\lib\site-packages\pydantic\decorator.cp38-win_amd64.pyd in pydantic.decorator.ValidatedFunction.call()

~\.conda\envs\hydra-zen\lib\site-packages\pydantic\decorator.cp38-win_amd64.pyd in pydantic.decorator.ValidatedFunction.init_model_instance()

~\.conda\envs\hydra-zen\lib\site-packages\pydantic\main.cp38-win_amd64.pyd in pydantic.main.BaseModel.__init__()

ValidationError: 1 validation error for F
x
  ensure this value is greater than 0 (type=value_error.number.not_gt; limit_value=0)

During handling of the above exception, another exception occurred:

TypeError                                 Traceback (most recent call last)
C:\Users\RYANSO~1\AppData\Local\Temp/ipykernel_15996/1076775370.py in <module>
----> 1 instantiate(conf, x=-2)

c:\users\ryan soklaski\hydra_utils\src\hydra_zen\_hydra_overloads.py in instantiate(config, *args, **kwargs)
    124     >>> instantiate(config, (1, 2, 3))  # static analysis can deduce that the result type is `list`
    125     [1, 2, 3]"""
--> 126     return hydra_instantiate(config, *args, **kwargs)
    127 
    128 

~\.conda\envs\hydra-zen\lib\site-packages\hydra\_internal\instantiate\_instantiate2.py in instantiate(config, *args, **kwargs)
    178         _convert_ = config.pop(_Keys.CONVERT, ConvertMode.NONE)
    179 
--> 180         return instantiate_node(config, *args, recursive=_recursive_, convert=_convert_)
    181     else:
    182         raise InstantiationException(

~\.conda\envs\hydra-zen\lib\site-packages\hydra\_internal\instantiate\_instantiate2.py in instantiate_node(node, convert, recursive, *args)
    247                         )
    248                     kwargs[key] = _convert_node(value, convert)
--> 249             return _call_target(_target_, *args, **kwargs)
    250         else:
    251             # If ALL or PARTIAL non structured, instantiate in dict and resolve interpolations eagerly.

~\.conda\envs\hydra-zen\lib\site-packages\hydra\_internal\instantiate\_instantiate2.py in _call_target(_target_, *args, **kwargs)
     62         return _target_(*args, **kwargs)
     63     except Exception as e:
---> 64         raise type(e)(
     65             f"Error instantiating '{_convert_target_to_string(_target_)}' : {e}"
     66         ).with_traceback(sys.exc_info()[2])

~\.conda\envs\hydra-zen\lib\site-packages\pydantic\error_wrappers.cp38-win_amd64.pyd in pydantic.error_wrappers.ValidationError.__init__()

TypeError: __init__() takes exactly 3 positional arguments (2 given)

```

which sucks... It still gets the job done, and, ironically, it raises a `TypeError` so 🤫 . But in all seriousness..

There are two solutions to this that I can think of:
1. See if pydantic is willing to modify `ValidationError` to accommodate this standard re-raising usage.
2. Handle catch the `ValidationError` in `hydra_zen.funcs.zen_processing`. But I *really* don't want to do this.

I will try to make 1 happen.

## Customizing validation behavior

pydantic allows their validator [to be configured](https://pydantic-docs.helpmanual.io/usage/validation_decorator/#custom-config). Accordingly, `validates_with_pydantic` can be passed a configured instance of `validate_arguments`.

```python
from pydantic import validate_arguments
CustomVal = builds(
    validates_with_pydantic,
    zen_partial=True,
    validator=builds(validate_arguments, config=dict(arbitrary_types_allowed=False)),
)

# use CustomVal as a zen-wrapper
```

our default configuration sets `dict(arbitrary_types_allowed=True)`, which seems like it is the more intuitive baseline configuration to me. Maybe there is a good reason to go with their default of `False`, but... it seems like that would collide with a lot of annotations, e.g. in pytorch-lightning!

## Why Ship Our Own Wrapper at All?

If from `pydantic.validate_arguments`, why bother with `validates_with_pydantic`? 

Because:
1.  applying `pydantic.validate_arguments` to a class-object doesn't work
2.  applying `validate_arguments` to an object multiple times breaks it... I should submit a fix to `pydantic`.

If they add support / fix *both* of these things, then we can probably just delete `validates_with_pydantic`

### Finding a Home for This  (Edit: moved!)
I don't really know where the right home is for this, within hydra-zen. `hydra_zen.experimental.third_party.pydantic` seems somewhat appropriate, and people will probably expect that they have to download a new third-party dependency to use it. 

One downside is that the yamls get to be unsightly due to the path:

```
_zen_target: builtins.dict
_zen_wrappers: hydra_zen.third_party.pydantic.validates_with_pydantic
```

but maybe we can get tricky with some path-shortening stuff? I don't really want to mess with it.. but we could maybe finess:

```
_zen_target: builtins.dict
_zen_wrappers: hydra_zen.validates_with_pydantic
```

I am definitely open to suggestions. Maybe drop experimental? `hydra_zen.third_party.pydantic.validates_with_pydantic`?

### Other thoughts

Anything stick out here? Any oversights on my part? How does the docstring read? Does this seem cool? Obscure?


rsokl:
Added much stronger tests for all type-validators. Will move away from `experimental`

rsokl:
🥳

PR #158, opened 2021-11-08
Docs: fix bad reference in pydantic docs

rsokl:


rsokl:
The failing pyright job looks to me to be a false positive. Tracking here: https://github.com/microsoft/pyright/issues/2549

PR #163, opened 2021-11-17
Add strict validation against Hydra's supported primitives

### LINKED ISSUE #149: Complex Numbers
So `builds` supports instantiating complex numbers:

```python
builds(list, [1 + 1j])
#or
builds(torch.tensor, 1+1j)
```

but omegaconf does not support serializing to yaml:

```python
to_yaml(builds(list, [1+1j])
```
leads to the following error
```
UnsupportedValueType: Value 'complex' is not a supported primitive type
    full_key: _args_[0][0]
    object_type=list
```

I'm not sure if we can add support for this, but should at least be documented.

rsokl:
The Eyes of T.J. Eckleburg, The Great Gatsby (2013)
> The Eyes of T.J. Eckleburg, *The Great Gatsby* (2013)

Closes #149 

## Adding strict validation against Hydra's supported primitives

Prior to this PR, hydra-zen would not check configured values to see if they are compatible with Hydra's supported primitives. This means that you could construct a config via `builds` and `make_config` that are destined to raise whenever Hydra tries to serialize or instantiate the config. This is no good!

**Before**:

```python
from hydra_zen import make_config, to_yaml, instantiate, builds

class MyCustomClass:
    pass
```

```python
>>> not_supported = MyCustomClass()
>>> Conf = make_config(a=not_supported)  # hydra-zen is OK with this even though it is doomed to fail
>>> to_yaml(Conf)   # will raise
>>> instantiate(Conf) # will raise
```

Also, our annotations did not provide any feedback along these lines

image

**After**:

Now `builds`, `make_config`, and `hydrated_dataclass` strive to strictly validate all configured values -- whether they are specified by the user or auto-populated from an object's signature. The goal here is to catch mistakes in configs, which would eventually be "snagged" by omegaconf/Hydra, before you even launch your app.

```python
>>> not_supported = MyCustomClass()
>>> Conf = make_config(a=not_supported) 
HydraZenUnsupportedPrimitiveError         Traceback (most recent call last)
HydraZenUnsupportedPrimitiveError: Building: dict ..
 The configured value <__main__.MyCustomClass object at 0x000001E01B72B4C0>, for field `a`, is not supported by Hydra. Serializing or instantiating this config would ultimately result in an error.
Consider using `hydra_zen.builds` to create a config for this particular value.
```

I hope that the error message is sufficiently clear

> HydraZenUnsupportedPrimitiveError: Building: dict ..
 The configured value <__main__.MyCustomClass object at 0x000001E01B72B4C0>, for field `a`, is not supported by Hydra. Serializing or instantiating this config would ultimately result in an error.
Consider using `hydra_zen.builds` to create a config for this particular value.

We even peek inside containers and validate their contents

```python
>>> builds(dict, a=[1, not_supported])  # will raise  (inside list)
>>> make_config(a={1: not_supported}))  # will raise  (as value in dict)
>>> make_config(a={not_supported: "a"}))  # will raise  (as key in dict)
```

Note that *all* of our type-checking used for this validation should adhere to the form: `type(x) is in SUPPORTED_TYPE`, and **not** `isinstance(x, SUPPORTED_TYPES)` because we do not handle/permit subclasses of valid primitive types.

## Providing Support for Additional Primitives

This PR isn't just about narc-ing on bad configs, it also adds new capabilities! We can add support for common types that Hydra doesn't handle. Presently, we add support for:

- bytes
- bytearray
- complex
- Counter
- deque
- Path
- PosixPath
- WindowsPath
- range
- set
- frozenset

```python
from pathlib import Path
from hydra_zen import to_yaml

def pp(x): print(to_yaml(x))
```

```python
>>> pp(make_config(a=Path.home()))
a:
  _args_:
  - C:\Users\Ryan Soklaski
  _target_: pathlib.Path

>>> instantiate(make_config(a=Path.home()))
{'a': WindowsPath('C:/Users/Ryan Soklaski')}

>>> pp(make_config(a={1, 2, 3}))
a:
  _target_: builtins.set
  _args_:
  - - 1
    - 2
    - 3

>>> pp(make_config(a=1+2j))
a:
  real: 1.0
  imag: 2.0
  _target_: builtins.complex
```

We're pretty fancy here - we even permit things like sets of complex numbers!

```python
>>> pp(make_config(a={1+2j, -10j}))
a:
  _target_: builtins.set
  _args_:
  - - real: -0.0
      imag: -10.0
      _target_: builtins.complex
    - real: 1.0
      imag: 2.0
      _target_: builtins.complex
```

## Improved Annotations

The annotation for our config-creating functions are much better now; type-checkers will now flag bad inputs. The following reflects the behavior of pyright/pylance under the "basic" type-checking mode.

image

### Exception vs Warning (for now)

Given that we raise an exception for this validation, upgrading from v0.3 to v0.4 could be a rude awakening for folks who are, for some reason, using hydra-zen to a bunch of ultimately-invalid configs.

I have been considering having this PR raise a warning on bad-configurations, and note that in a future version the warning will become an exception. The reasons why I opted for going the more abrupt route of raising an error is:
1. For current user code for which this PR will introduce exceptions, I expect that their code would be raising *already*, but just downstream. I suppose that that folks *could* be using `builds` and `make_config` in a way that is detached from Hydra... but that really isn't what `hydra_zen`  is for, and I wouldn't be surprised if literally no one is doing this.
2. It is harder to track down the source of a warning than it is the emitter of an exception. So going the route of raising an exception actually makes it easier for new users to quickly diagnose the source of their error.
3. I want people to immediately be able to incorporate this improved validation in the automated tests. Converting warnings to exceptions on the user-side is way too much work.

The above is my reasoning on this topic. I am open to discussion if there are other opinions.

### Any Other Low-Hanging Fruit for Supporting Primitives?

I plan to add support for `numpy.ndarray` and `torch.Tensor`, but in a separate PR because handling the dynamic imports is a bit trickier. 

That all being said, are there other primitives, beyond `set`, `frozenset`, `complex`, and `pathlib.Path`, that we should add support for? For example, `pathlib.Path` jumped out at me as being *super* useful. Obviously there are plenty of fish in the std-lib sea; support can be added incrementally.

### Possible Future Capability: Enable Users to Register Their Own Primitives / Conversion Strategies

There are various ways by which we could enable users to register new types and corresponding conversion strategies. 

The biggest "Pro" is that if someone is regularly using some custom type in their config, registering a converter makes config-design *much* more elegantly.

The biggest "Con" that I can think of is that this can introduce "secret" dependencies on dynamic code. E.g. A user's config-generating code can only be run *after* that converter has been registered. Fortunately, the resulting config (either the dataclass or yaml) should be independent of this dependency. This is an improvement of omegaconf's capability for registering resolvers -- where even yamls can become secretly tethered to third-party code.

# Implementation Details

## Restricted Primitives for Dictionary Keys
omegaconf does not permit dataclasses as keys in dictionaries, even if they would instantiate to a hashable value. Thus the extended primitive support is not provided for dict-keys

image

## No Support for `attrs` (for now)

We currently do not provide any support for the [attrs](https://www.attrs.org/en/stable/) library; it is likely that `attrs` instances will get caught by our validation in its current form. This is not likely to impact most hydra-zen users in the near term, but we should eventually add support for it.

## Support for Enum

Although Omegaconf provides support for `Enum`, the yaml serialization process only retains the enum-member's name:

```python
from enum import Enum

from hydra_zen import make_config, builds, to_yaml, instantiate
from omegaconf import OmegaConf, ListConfig, DictConfig

class Color(Enum):
    red = 1
    green = 2
```
 ```python
>>> OmegaConf.create(to_yaml({1: Color.red}))
{1: 'red'}
```

It appears that the enum-member can only be restored if a schema is provided that contains the `Enum` subclass in the annotation. Thus hydra-zen instead provides specialized support for `Enum`:

```python
>>> OmegaConf.create(to_yaml(make_config(a={1: Color.red})))
{'a': {1: {'_target_': '__main__.Color', '_args_': [1]}}}
```

this way, one can roundtrip from a yaml and restore the actual enum-member:

```python
>>> instantiate(OmegaConf.create(to_yaml(make_config(a={1: Color.red}))))
{'a': {1: <Color.red: 1>}}
```

### DIFF
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -110,19 +110,7 @@ def _convert_set(value: set) -> Type[Builds[Type[Set]]]:
     return builds(set, tuple(value))
 
 
-@dataclass
-class _ConfigComplex:
-    real: Any
-    imag: Any
-    _target_: str = _utils.get_obj_path(complex)
-
-
-def _convert_complex(value: complex) -> Builds[Type[complex]]:
-    return cast(Builds[Type[complex]], _ConfigComplex(real=value.real, imag=value.imag))
-
-
 ZEN_VALUE_CONVERSION[set] = _convert_set
-ZEN_VALUE_CONVERSION[complex] = _convert_complex
 
 
 def _get_target(x):
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -1454,7 +1454,7 @@ def builds(
                     field_.default,
                     allow_zen_conversion=False,
                     error_prefix=_utils.building_error_prefix(target),
-                    field_name=field.name + " (set via inheritance)",
+                    field_name=field_.name + " (set via inheritance)",
                 )
             del field_
 
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -110,7 +110,12 @@ def _convert_set(value: set) -> Type[Builds[Type[Set]]]:
     return builds(set, tuple(value))
 
 
+def _convert_frozenset(value: set) -> Type[Builds[Type[FrozenSet]]]:
+    return builds(frozenset, tuple(value))
+
+
 ZEN_VALUE_CONVERSION[set] = _convert_set
+ZEN_VALUE_CONVERSION[frozenset] = _convert_frozenset
 
 
 def _get_target(x):
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -113,8 +113,6 @@ def _cast_via_tuple(dest_type: Type[_T]) -> Callable[[_T], Type[Builds[Type[_T]]
     return converter
 
 
-a = _cast_via_tuple(set)
-
 ZEN_VALUE_CONVERSION[set] = _cast_via_tuple(set)
 ZEN_VALUE_CONVERSION[frozenset] = _cast_via_tuple(frozenset)
 ZEN_VALUE_CONVERSION[deque] = _cast_via_tuple(deque)
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -521,6 +521,9 @@ def create_just_if_needed(value: _T) -> Union[_T, Type[Just]]:
     return value
 
 
+_KEY_ERROR_PREFIX = "Configuring dictionary key:"
+
+
 def sanitize_collection(x: _T) -> _T:
     """Pass contents of lists, tuples, or dicts through sanitized_default_values"""
     type_x = type(x)
@@ -528,7 +531,11 @@ def sanitize_collection(x: _T) -> _T:
         return type_x(sanitized_default_value(_x) for _x in x)  # type: ignore
     elif type_x is dict:
         return {
-            sanitized_default_value(k): sanitized_default_value(v)
+            # Hydra doesn't permit structured configs for keys, thus we only
+            # support its basic primitives here.
+            sanitized_default_value(
+                k, allow_zen_conversion=False, error_prefix=_KEY_ERROR_PREFIX
+            ): sanitized_default_value(v)
             for k, v in x.items()  # type: ignore
         }
     else:
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -43,7 +43,7 @@ from hydra_zen.errors import (
 )
 from hydra_zen.funcs import get_obj, partial, zen_processing
 from hydra_zen.structured_configs import _utils
-from hydra_zen.typing import Builds, Importable, Just, PartialBuilds
+from hydra_zen.typing import Builds, Importable, Just, PartialBuilds, SupportedPrimitive
 from hydra_zen.typing._implementations import DataClass, HasPartialTarget, HasTarget
 
 from ._value_conversion import ZEN_SUPPORTED_PRIMITIVES, ZEN_VALUE_CONVERSION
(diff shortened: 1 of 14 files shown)


jgbos:
>HydraZenUnsupportedPrimitiveError: Building: dict ..
The configured value <main.MyCustomClass object at 0x000001E01B72B4C0>, for field a, is not supported by Hydra. Serializing or instantiating this config would ultimately result in an error.
Consider using hydra_zen.builds to create a config for this particular value.

I'm not sure this is clear enough for beginners. Do you mean that a user should try to build a configuration for the value rather than directly using the value, e.g., `builds(MyCustomClass)` instead?  Or in the (before this PR) world when I needed to create a special class for `complex` types? 

Maybe just reorder the sentence, "Consider creating a configuration for the particular value using hydra_zen.builds".

jgbos:
Your logic on the the "exception vs. warning" is good.  

I can only think of use case that might benefit from warnings instead: what if the user doesn't use the Hydra backend but wants to use the configs? 

I know, not a strong argument against exceptions, just trying to think of any good reason to go to warnings.

jgbos (commented):
Lots of changes, nothing pops out as unusual.  Excited to have complex support :)

rsokl:
I just realized that we should include omegaconf containers amoung the supported primitives.

src/hydra_zen/structured_configs/_implementations.py
```diff
@@ -485,32 +515,94 @@ def create_just_if_needed(value: _T) -> Union[_T, Type[Just]]:
...
+    # we don't use isinstance because we don't permit subclasses of supported
+    # primitives
+    if allow_zen_conversion and type_value in ZEN_SUPPORTED_PRIMITIVES:
+        type_ = type(resolved_value)
+        conversion_fn = ZEN_VALUE_CONVERSION.get(type_)
+
+        if conversion_fn is not None:
+            return conversion_fn(resolved_value)
+
+    if type_value in HYDRA_SUPPORTED_PRIMITIVES or is_dataclass(resolved_value):
```
rsokl, line 554: We need to add support for:
- subclasses of Enums
- omegaconf objects

jgbos: Oh Enums, that's a good one as we have an Enum class in our radar sim.  I'll look through my code and see if there any other good primitives.

jgbos (approved):
Just used the enum support for some configs :)

rsokl:
Finally wrapping up on this. I added some additional discussion and remarks on implementation details in the original post.

PR #175, opened 2021-12-05
Loosen type annotations to avoid false-positives on mutable generic containers

### LINKED ISSUE #173: Inconsistent static analysis via pyright
```python
x = [1, 2, 3]
make_config(a=[1, 2, 3])
make_config(a=x)  # pyright marks this as invalid based on our annotations
```

This seems like a pyright issue, but we might consider revising annotations to accommodate

rsokl:
Closes #173 

Broadens our type-annotations for mutable containers (dicts, sets, lists, deques) to avoid the risk of users hitting false-positives in their code. ~On the downside, we can no longer provide precise feedback about whether, e.g., a list has OK contents for Hydra via our annotations.~ (EDIT: See [comment](https://github.com/mit-ll-responsible-ai/hydra-zen/pull/175#issuecomment-986170971)).  At runtime we still provide accurate checking.

### DIFF
--- a/src/hydra_zen/typing/_implementations.py
+++ b/src/hydra_zen/typing/_implementations.py
@@ -7,14 +7,11 @@ from pathlib import Path
 from typing import (
     Any,
     Callable,
-    Counter,
     Deque,
     Dict,
     FrozenSet,
     Generic,
-    List,
     NewType,
-    Set,
     Tuple,
     TypeVar,
     Union,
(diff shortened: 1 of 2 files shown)


Jasha10:
Not sure if this would work, but one idea that came to mind is to use immutable interfaces, i.e. `Sequence[SupportedPrimitive]` instead of `List[SupportedPrimitive]` and `Mapping[...]` instead of `Dict[...]`.

rsokl:
> one idea that came to mind is to use immutable interfaces,

Yeah I think this is a good call. I wasn't clearly about the pros and cons here. Using `Sequence[...]` and `Mapping[...]` where:
- pro: provides high-quality static checking of the *contents* of collections
- con: is overly permissive of the kinds of sequences/mappings permitted

Versus using `list` and `dict`, where:
- pro: it correctly tells users to not use unsupported containers
- cons: provides *no* feedback about the contents of the containers

Given that users will rarely use containers *other* than list/tuple/dict, I think going with `Sequence[...]` and `Mapping[...]` will provide much more valuable information without permitting too many false negatives, and with no false positives (at least I think!)

PR #191, opened 2022-01-09
Don't enable support for _partial_ until we can begin running CI against it

rsokl:
We implemented future support for `_partial_` in Hydra's instantiate API, but we should not permit this to be toggled "on" on the user's end until we can actually test our support in our CI.


Jasha10:
I'll try to get a nightly build running on Hydra/OmegaConf as soon as I can.

rsokl:
No rush! I am thinking that we might release `0.5.0` pretty soon, given our recent improvements to sig-parsing and type-widening. And I realized that if there is something wrong with our support for `__partial__` then it could cause a real headache for `0.5.0` users who later upgrade Hydra.

PR #210, opened 2022-01-22
Improve landing page docs

rsokl:
Documentation is such a fine balance between being concise and being informative 😅 

I think that the [landing page for our docs](https://mit-ll-responsible-ai.github.io/hydra-zen/) is too vague. For those who aren't familiar with Hydra, they are probably scratching their head and wondering what the heck Hydra exactly *is*. For those who are familiar with Hydra, they might not be able to glean what the difference is between Hydra and hydra-zen.

I also think that we can do a better job highlighting some bells and whistles of hydra-zen, to entice more experienced Hydra users.

Hopefully this update to our docs helps to remedy these issues!

~[Here are instructions for building our docs](https://github.com/mit-ll-responsible-ai/hydra-zen/tree/main/docs). I wish there was an easy way for me to host these changes, for easy viewing.~

You can download the built docs from here:
[hydra-zen-docs.zip](https://github.com/mit-ll-responsible-ai/hydra-zen/files/7919095/hydra-zen-docs.zip)

@jgbos  it would be great if you could review this.

@Jasha10 it would be awesome to get your feedback on this, if possible. No worries _at all_ if you do not have time!


Jasha10:
The new landing page gives an introduction to the Hydra+Zen stack that is gentle enough for neophytes, which is great! For experienced Hydra users, I feel that it's hard to see through the Hydra part and find part that is Zen.

My sense is that, for most Hydra users, the vast majority of configs are written in YAML. Dataclasses are a know option, but they are seen as too cumbersome to be used at scale. The fact that Hydra-Zen automates the production of structured configs is something that, in my opinion, could not be made too obvious.

My biased recommendation is that the section about "Differences Between Hydra and hydra-zen" be shorter, pithier, and closer to the top of the page. The dataclass-generation mechanism, which is the crown jewel of hydra-zen, should be something that a sleep-deprived and over-caffeinated Hydra user can pick up on within the first 5 seconds.

There is a tradeoff here: if you mention dataclass generation out of the blue near the top of the webpage, there's a chance you'll scare off those who have no experience with Hydra. On the flipside, if this feature is not made obvious enough, experienced Hydra users will have trouble figuring out what use cases are enabled by Hydra-Zen that are not covered by Hydra itself. Such use cases are mentioned in the abstract near the top of the page, but a concrete example is what will really close the sale.

Here is the pitch that I would give to an experienced Hydra user:

---

Hydra-Zen automates the generation of structured configs!
<table>

<tr>
<td> Using Plain Hydra </td> <td> Using Hydra-Zen </td>
</tr>

<tr>
<td>

```python
from dataclasses import MISSING, dataclass, field
from hydra_zen import builds

def foo(bar: int, baz: list[str], qux: float = 1.23):
    ...

@dataclass
class FooConf:
    _target_: str = "__main__.foo"
    bar: int = 2
    baz: list[str] = field(default_factory=lambda: ["abc"])
    qux: float = 1.23
```
<td/>

```python
from hydra_zen import builds

def foo(bar: int, baz: list[str], qux: float = 1.23):
    ...

FooConf = builds(foo, bar=2, baz=["abc"], populate_full_signature=True)



```
</tr>

</table>

image

---

The example is completely self-contained, and the side-by-side nature makes the comparison super obvious.

Anyway, that's my two cents; take it or leave it :)

rsokl:
This feedback is very useful. Thank you for taking the time to write this up, @Jasha10 ! 

> The fact that Hydra-Zen automates the production of structured configs is something that, in my opinion, could not be made too obvious.
>...
> if this feature is not made obvious enough, experienced Hydra users will have trouble figuring out what use cases are enabled by Hydra-Zen that are not covered by Hydra itself.

I totally agree, and it is helpful for me to learn that this is your perspective as well. My experience of seeing new-to-Hydra readers bounce off of my earlier docs made me shy about being too detailed about `builds` et al early on, but at a tradeoff -- as you correctly pointed out.

I am sure I can find the proper messaging to warn new readers. and then to draw the attention of Hydra users to a pitch similar to the one that your wrote out. I am thinking I might have a section, that follows immediately after "Installation", called "What Does hydra-zen Add to Hydra?", where I explained to Hydra veterans our ability to generate structured configs. And then I can provide a shortened version of "Hydra and hydra-zen at a Glance" for total newcomers.

Jasha10:
Sounds good!

rsokl:
[hydra-zen-docs.zip](https://github.com/mit-ll-responsible-ai/hydra-zen/files/7942943/hydra-zen-docs.zip)

Attached is the latest build of the docs. I decided to:
- Remove both "What Does hydra-zen Add to Hydra?" and "hydra-zen at a Glance" altogether
- Add an "Attention, Hydra users:" call-out box that basically uses @Jasha10 's pitch exactly

After getting feedback from additional users, the consensus was that the content I added in this PR was just too lengthy and dense. The pitch that @Jasha10  laid out is really what needs to get across at a glance. From there, newbie readers can click through our tutorials to learn about the basics of Hydra and hydra-zen.

It would still be useful for me to add a page in our "Explanations" section that provides an overview of Omegaconf, Hydra, and hydra-zen all in one place. I also got feedback that it would be nice to show a `argparse` vs Hydra+hydra-zen comparison somewhere.

jgbos (approved):

rsokl:
Thanks @Jasha10  and @jgbos  for the help on this!

Issue #205, opened 2022-01-19
Calling methods via instantiate API


rsokl:
This is just a prototype. We can use zen-wrappers to let people specify a method call via instantiate:

```python
from hydra_zen import builds, instantiate

def outer(Target, *, mthd_name, mthd_args, mthd_kwargs):
    
    def wrapper(*args, **kwargs):
        return getattr(Target(*args, **kwargs), mthd_name)(*mthd_args, **mthd_kwargs)
    return wrapper

def call_method(mthd_name, *mthd_args, **mthd_kwargs):
    return builds(
        outer,
        mthd_args=mthd_args,
        mthd_name=mthd_name,
        mthd_kwargs=mthd_kwargs,
        populate_full_signature=True,
        zen_partial=True
    )
```

```python
class A:
    def __init__(self, x):
        self.x = x
    def x_plus_y(self, y):
        return self.x + y
```

```python
>>> instantiate(builds(A, x=-11, zen_wrappers=call_method("x_plus_y", y=2)))
-9
```

I can't say I love this form factor, or the readability of this. Specifying the method via its name (a string) is not very zen. I would want to see if I can make the yaml more legible:
```
_target_: hydra_zen.funcs.zen_processing
_zen_target: __main__.A
_zen_wrappers:
  _target_: hydra_zen.funcs.zen_processing
  _zen_target: __main__.call_method
  _zen_partial: true
  mthd_name: x_plus_y
  'y': 2
x: -11
```

Some things for us to consider:
- Would it be worthwhile for us to provide users with this sort of wrapper?
- Can the wrapper be designed to handle class & instance & static methods?
- Can we avoid requiring the method be specified by name (string)?
- Is there some sort of optional validation API that we can expose for zen-wrappers in general? So that `builds(A, x=-11, zen_wrappers=call_method("x_plus_y", y=2))` can validate that the method call will be valid upon instantiation?

jgbos:
This doesn't work with the code above:
```python
>>> instantiate(builds(torch.Generator, zen_wrappers=call_method("manual_seed", 42)))
...
TypeError: Error instantiating 'hydra_zen.funcs.zen_processing' : outer() got multiple values for argument 'mthd_name'
```

rsokl:
Ah, whoops. Yeah things get a little complicated with `*args` and partial. I updated the implementation in the top post, there should no longer be an issue.

jgbos:
[...]

Issue #93, opened 2021-09-09
New feature: `hydra_zen.builds_decorator`


jgbos:
We had a discussion on Teams about possibly adding a `builds_decorator` of the form:

```python
def builds_decorator(f):
    Conf = builds(f, populate_full_sig=True)
    f.config = Conf
    return f
```

Then we can automatically build configs for a function or class.  For example, to get the config for the following function,

```python
@builds_decorator
def foo(x, val=1):
  return x, val
```

simply access the `config` attribute,

```python
FooConfig = foo.config(val=2)
```

rsokl:
I like this idea and have been trying to think of ways to make `builds` a bit more ergonomic along these lines. That being said, it feels a bit "risky" to actually ship this:

- IDEs will complain about accessing `.config(...)`

![image](https://user-images.githubusercontent.com/29104956/132761853-a0916a54-cc83-4572-8198-07087bf8b80b.png)

- There is the potential of colliding with objects that already have a `config` method.
- Getting the type-annotations right seem hard (maybe impossible)
- Mutating an object to have a new, hydra-zen specific method could lead to some unforeseen issues / probably isn't considered good practice
- Once we ship this, removing it later would break peoples experiments pretty badly

Maybe we could simply document this as a recipe for library authors to use, but they are ultimately responsible for opting-in/maintaining it. Like, I can definitely imagine `mushin` implementing this.

jgbos:
Good point.  I just remember we were talking about so thought I'd document our discussion at least :)

rsokl:
Yep! I think it is good to document our thoughts on these sorts of things. It can be also helpful for us to reference these issues if users request similar features later.

rsokl:
@jgbos what do you think about this approach

```python
from functools import partial


def as_builds(target):
    return partial(builds, target, populate_full_signature=True)


RandomCrop = as_builds(transforms.RandomCrop)
RandomHorizontalFlip = as_builds(transforms.RandomHorizontalFlip)
ColorJitter = as_builds(transforms.ColorJitter)
RandomRotation = as_builds(transforms.RandomRotation)
ToTensor = as_builds(transforms.ToTensor)
Normalize = as_builds(transforms.Normalize)
Compose = as_builds(transforms.Compose)

Cifar10TrainTransformsConf = Compose(
    transforms=[
        RandomCrop(size=32, padding=4),
        RandomHorizontalFlip(),
[...]

PR #219, opened 2022-02-04
Implements prototype of hydra_zen.like

rsokl:
Potentially addresses: #205

#219  prototypes a potential solution to the above, that is much more powerful than what I had initially proposed. This is inspired by @jgbos ' [out of left field](https://github.com/mit-ll-responsible-ai/hydra-zen/issues/205#issuecomment-1017951452) idea 😄 

Let's see `hydra_zen.like` in action:

```python
>>> from hydra_zen import just, instantiate, like, to_yaml
>>> import torch

>>> GenLike = like(torch.Generator)  # 'memorizes' sequence of interactions with `torch.Generator`
>>> seeded = GenLike().manual_seed(42)
>>> seeded   
Like(<class 'torch._C.Generator'>)().manual_seed(42)

>>> Conf = just(seeded)  # creates Hydra-compatible config for performing interactions via instantiation

>>> print(to_yaml(Conf))
_target_: hydra_zen.structured_configs._implementations.make_like
_args_:
- torch._C.Generator
- - - __call__
    - []
    - {}
  - manual_seed
  - - __call__
    - - 42
    - {}

>>> generator = instantiate(Conf)  # calls torch.Generator().manual_seed(42)
<torch._C.Generator at 0x14efb92c770>
```

what is going on here?

`hydra_zen.like(<target>)` returns a `_Tracker` instance that records all subsequent interactions with `<target>`. Right now, "interactions" are restricted to accessing attributes and calling the target. `hydra_zen.just` then knows how to convert a `_Tracker` instance into a valid Hydra config.

Let's use this to make a config that actually returns the `manual_seed` method of our Generator-instance:

```python
>>> manual_seed_fn = GenLike().manual_seed

>>> instantiate(just(manual_seed_fn))
<function Generator.manual_seed>
```

Here is a silly example, just to help give more of a feel for this:

```python
>>> listy = like(list)
>>> x = listy([1, 2, 3])
>>> y = x.pop()

# note that `x` *does not* reflect any mutation
>>> instantiate(just(x)), instantiate(just(y))
([1, 2, 3], 3)
```

## Current thoughts on this prototype

One nice thing about `like` is that looks like a simple pass-through to IDEs, so you get auto-completes, static type checks, etc. as you build your "like expression" prior to creating the config.

Some obvious to-dos would include:

- Adding as much validation as possible. E.g. `GenLike().manual_seedling(42)` should raise a runtime error because `manual_seedling` is not an attribute of `Generator`.
- Making the yaml-representation of the `just(like(<...>).<...>)` expression as clean as possible 
   - doing `just(like(<target>)(*args, **kwargs))` should reduce down to `builds(<target>, *args, **kwargs)`
- Enable users to directly provide `like`'d objects in their configs without them needing to call `just` first.

I would definitely like to get feedback on this. I only just thought of it, and will have to think carefully about potential pitfalls as well as about possible use cases that I haven't anticipated yet.

## Additional functionality

`like` could also fulfill the role of [the proposed `hydra_zen.hydrate`](https://github.com/mit-ll-responsible-ai/hydra-zen/issues/93)

```python
import torch as tr

tr = like(tr)

tr.tensor([1., 2.])  # is equivalent to `like(tr.tensor)([1.0, 2.0])`
```

### DIFF
--- a/src/hydra_zen/__init__.py
+++ b/src/hydra_zen/__init__.py
@@ -9,6 +9,7 @@ from ._hydra_overloads import (
     to_yaml,
 )
 from ._launch import launch
+from ._like import like
 from ._version import get_versions
 from .structured_configs import (
     ZenField,
@@ -36,6 +37,7 @@ __all__ = [
     "ZenField",
     "make_custom_builds_fn",
     "launch",
+    "like",
 ]
 
 __version__ = get_versions()["version"]
--- a/src/hydra_zen/_like.py
+++ b/src/hydra_zen/_like.py
@@ -1,0 +1,41 @@
+from typing import TypeVar, cast
+
+T = TypeVar("T")
+
+
+class _Tracker:
+    def __init__(self, origin, tracks=None):
+        self.origin = origin
+
+        self.tracker = [] if tracks is None else tracks.copy()
+
+    def __repr__(self) -> str:
+        base = f"Like({repr(self.origin)})"
+        for item in self.tracker:
+            if isinstance(item, tuple):
+                _, args, kwargs = item
+                contents = ""
+                if args:
+                    contents += ", ".join(repr(x) for x in args)
+                if kwargs:
+                    if args:
+                        contents += ", "
+                    contents += ", ".join((f"{k}={v}" for k, v in kwargs.items()))
+
+                base += f"({contents})"
+            else:
+                base += f".{item}"
+        return base
+
+    def __call__(self, *args, **kwargs):
+        return _Tracker(self.origin, self.tracker + [("__call__", args, kwargs)])
+
+    def __getattr__(self, name):
+        # IPython will make attribute calls on objects; we don't want to track these.
+        if name != "_ipython_canary_method_should_not_exist_" and name != "__wrapped__":
+            return _Tracker(self.origin, self.tracker + [name])
+        return self
+
+
+def like(obj: T) -> T:
+    return cast(T, _Tracker(obj))
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -60,6 +60,7 @@ from hydra_zen.typing import (
 )
 from hydra_zen.typing._implementations import DataClass, HasTarget, _DataClass
 
+from .._like import _Tracker
 from ._value_conversion import ZEN_VALUE_CONVERSION
 
 _T = TypeVar("_T")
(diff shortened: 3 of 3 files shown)


addisonklinke:
This sounds intriguing, but I think I need a bit more time to wrap my head around exactly what's going on (it probably doesn't help that I haven't played around with `just` either). Based on this part of your example

```
generator = instantiate(Conf)  # calls torch.Generator().manual_seed(42)
```

I'm understanding that the combination of `like` and `just` enables `instantiate` to both initialize a `torch.Generator` instance _and_ call the instance method `manual_seed` in a single line. Whereas previously this pattern would've required multiple steps 

```
>>> GenConf = builds(torch.Generator)
>>> print(to_yaml(GenConf))
_target_: torch.Generator

>>> generator = instantiate(GenConf)
>>> generator.manual_seed(42)  # Can't happen until the above line runs
```

So in a way, it's like the recursive instantiation added in Hydra 1.1 except instead of recursing into the init parameters of the target object, we're recursing into the target object's methods and attributes _once it's already_ instantiated.  Is that the advantage you're looking to demonstrate, or am I missing something?

Could you elaborate on how this fulfills the role of `hydrate`? In reading your [original proposal](https://github.com/mit-ll-responsible-ai/hydra-zen/issues/93#issuecomment-921282227), I thought that `hydrate` would take an imported module and return its corresponding Hydra config node (which then could be stored in the `ConfigStore` and exposed with the application's CLI interface). The torch example above makes it seem like there's a different design goal - that we should first `hydrate` the objects we'll use and then subsequent interactions are memorized. If that is the intent, is it supposed to tie into the CLI interface eventually? 

Thanks in advance for the clarification! I'll be interested to understand better

rsokl:
> I'm understanding that the combination of like and just enables instantiate to both initialize a `torch.Generator` instance and call the instance method manual_seed in a single line. Whereas previously this pattern would've required multiple steps.

That is correct. Without `like`, there is no easy way to write a config that, when instantiated, initializes a class *and then* calls a method. 

In general, given 

```python
conf = hydra_zen.like(obj)
```` 

`like` will record all subsequent "interactions" with `obj`, and will replay those interactions upon instantiation. E.g.


```python
conf = conf("hello")
conf = conf.a
conf = conf.b

out = instantiate(conf)  # performs: `obj("hello").a.b`
```

So the advantage that I am demonstrating is that `like` would enable us to express configs that are otherwise impossible / very cumbersome to write. And this could also be used to "record interactions" with a module object.

> Could you elaborate on how this fulfills the role of hydrate? In reading your https://github.com/mit-ll-responsible-ai/hydra-zen/issues/93#issuecomment-921282227, I thought that hydrate would take an imported module and return its corresponding Hydra config node (which then could be stored in the ConfigStore and exposed with the application's CLI interface).

I'm glad you asked this; the way you posed the question helped me realize that there is indeed a significant discrepancy between the proposed `hydrate` vs using `like` on a module object. You are correct that `hydrate` could return a Hydra config node for an entire module. `like`, on the other hand, would only help to make config-creation code look more like native Python.

I see now how `hydrate` is closer to a direct / generalized replacement to projects like `hydra-torch`, and that `like` would not facilitate this. I'll open an issue dedicated to `hydrate` to see if we can move forward with it.

rsokl:
Closing this for now. I think it is a neat idea, but if the main application of `like` is "instantiate + method call", then it is a bit overkill.

rsokl:
@addisonklinke  see #256 for a new discussion of `hydrate`

PR #272, opened 2022-05-06
Updates to `launch`

jgbos:
This PR does two things:

- Adds flag to convert a `dataclass` input to a `DictConfig`.  This is a fix and response to the cloudpickle [issue](https://github.com/cloudpipe/cloudpickle/issues/386)
- Refactors `launch` to comform with Hydra's Compose API.  Currently we cannot use `compose` as it does not support `muliturn`, but these changes make it easy to update if `compose` eventually supports `multirun`. 

@rsokl I added cloudpickle to some of the test requirements.  I'm not sure how to test without it.

### DIFF
--- a/src/hydra_zen/_launch.py
+++ b/src/hydra_zen/_launch.py
@@ -1,5 +1,6 @@
 # Copyright (c) 2022 Massachusetts Institute of Technology
 # SPDX-License-Identifier: MIT
+import warnings
 from dataclasses import fields, is_dataclass
 from typing import Any, Callable, List, Mapping, Optional, Type, Union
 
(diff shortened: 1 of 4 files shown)


src/hydra_zen/_launch.py
```diff
@@ -203,57 +205,84 @@ def launch(
...
+            # convert Dataclass to a DictConfig
+            cfg = OmegaConf.structured(config)
+            config = OmegaConf.create(OmegaConf.to_container(cfg))
+        else:
+            if len(fields(config)) == 0:
+                raise ValueError(
+                    """There is an issue with your dataclass.  If you previously executed with a
+                `hydra/launcher` that utilizes cloudpickle (e.g., hydra-submitit-launcher), there is a known
+                issue with dataclasses (see: https://github.com/cloudpipe/cloudpickle/issues/386). You will have
+                to restart your interactive environment.  To avoid this issue you can use the option `to_dictconfig=True`."""
```
rsokl, line 220: This check will prevent empty configs from working:

```python
from hydra_zen import make_config

from dataclasses import fields

>>> fields(make_config())
()
```

This would break some of our basic mushin workflow examples

rsokl: Actually, when you get rid of this check, it would be good to add a regression test to ensure that `launch` always supports an empty config.

rsokl: You could check the length of the fields before and after the run, and then raise a warning to the user.

jgbos: yeah, I was just thinking about that. I'm not sure we can have a check for this (maybe a warning but that could get irritating).

tests/test_launch/test_implementations.py
```diff
@@ -80,15 +79,7 @@ def task_fn(cfg):
...
-            with pytest.warns(
-                warnings.warn(
-                    "There may ben an issue with your dataclass.  If you just executed with a "
-                    + "`hydra/launcher` that utilizes cloudpickle (e.g., hydra-submitit-launcher), there is a known "
-                    + "issue with dataclasses (see: https://github.com/cloudpipe/cloudpickle/issues/386). You will have "
-                    + "to restart your interactive environment ro run `launch` again.  To avoid this issue you can use the option "
-                    + "`to_dictconfig=True`."
-                )
-            ):
+            with pytest.warns(UserWarning):
```
jgbos: ahh, that's how you do that.

PR #290, opened 2022-06-23
Remove restrictions on inheriting partial'd configs

### LINKED ISSUE #288: hydra-zen unable to extend hydra-zen functionality for base builder classes since at least version 0.7
Note: this worked fine in hydra-zen =< v0.5 and is a follow-on from this conversion: https://github.com/mit-ll-responsible-ai/hydra-zen/issues/284#issuecomment-1159009054

Let's say I wanted to make a torch optimizer and a bunch of its children:

```python
BuildsOptimizer = builds(torch.optim.Optimizer, zen_partial=True)
BuildsSGD = builds(
    torch.optim.SGD, 
    populate_full_signature=True,
    zen_partial=True,
    zen_meta={"optimizer": "SGD"},
    builds_bases=(BuildsOptimizer,),
)
ConfigStore.instance().store(...
```
This will fail to compile with the warning that hydra-zen functionality cannot be extended. Instead, `BuildsOptimizer` would have to look like this:
```python
BuildsOptimizer = builds(torch.optim.Optimizer)
BuildsSGD = ...
```

However, if I want type hinting in my `hydra_zen.ZenField` that explains how the field should be used, I'd do like this:
```python
optimizer_field = ZenField(name="optimizer", hint=BuildsOptimizer)
optimizer_default = [{"optimizer": "really_good_optimizer"}]
```
This feels in-congruent since my hint now no-longer explains that everything here really ought to be built with a partial- the base class requires the model weights and the model weights won't be known until runtime. The partial is really part of the abstraction of the base class and not an implementation detail of any of my optimizers.

P.S. love this library. Y'all rock!

### LINKED ISSUE #289: Eliminating restrictions on  `builds(..., zen_partial=<...>, builds_bases=(<...>))`
There are currently two restrictions that hydra-zen imposes on `builds(..., zen_partial=<...>, builds_bases=(<...>))`:

## 1. Inheriting from a partial-builds parent

Inheriting from a parent where `zen_partial=True` requires that all children explicitly specify `zen_partial=True`. 

```python
parent_with_partial = builds(int, zen_partial=True)

child = builds(int, builds_bases=(parent_with_partial,)) # Not allowed
child = builds(int, zen_partial=False, builds_bases=(parent_with_partial,)) # Not allowed

child = builds(int, zen_partial=True, builds_bases=(parent_with_partial,)) # Okay; zen_partial=True for child
```

This restriction is due to the fact that the default argument for `zen_partial` is `False`. Thus there is presently no way for us to distinguish between a user leaving `zen_partial` unspecified and explicitly specifying that they want `zen_partial=False` for the child.

### Solution

We will update the default of `zen_partial` to be `None` so that we can distinguish between `zen_partial` being left un-specified and it being set to `False` explicitly. With this change in place, the following behaviors will manifest:

```python
parent_with_partial = builds(int, zen_partial=True)

child = builds(int, builds_bases=(parent_with_partial,)) # zen_partial=True for child (via inheritance)

child = builds(int, zen_partial=None, builds_bases=(parent_with_partial,)) # zen_partial=True for child (via inheritance)

child = builds(int, zen_partial=False, builds_bases=(parent_with_partial,)) # zen_partial=False for child

child = builds(int, zen_partial=True, builds_bases=(parent_with_partial,)) # zen_partial=True for child
```

## 2. Mixing `_partial_=True` with zen-processing features

Hydra's added support for partial instantiation made possible the formation of configs like:

```
_partial_: true
_zen_target: SomeTarget
```

where instantiating this config will produce:

```
functools.partial(<function zen_processing at 0x0000024DA072FD30>, _zen_target=SomeTarget)
```

this is not desirable as the `.func` attribute of this partial'd object should return `SomeTarget`, it will instead return `hydra_zen.funcs.zen_processing`

Even more insidious is the case where we have:

```
_partial_: true
_zen_target: SomeTarget
_zen_partial: true
```

Here, instantiating this config will produce:

``` python
>>> out = instantiate(conf)
>>> out
functools.partial(<function zen_processing at 0x0000024DA072FD30>, _zen_target=SomeTarget, zen_partial=True)
>>> out()  # uh oh!
functools.partial(SomeTarget)
>>> out()()  # it was double partial'd!
SomeTarget
```

I.e. the presence of `_partial_=true` and `_zen_partial=true` means that the target will be *double partial'd*. For these reasons hydra-zen currently prohibits inheriting from a parent that uses zen-processing features into a child that is: a partial-builds and does not itself use zen-processing features.

Specifically this leads to the following restrictions:

```python
P = builds(int, zen_partial=True)  # has _partial_ = True
builds(int, zen_partial=true_or_false, builds_bases=(P,), zen_meta=dict(a=1))  # Not allowed
```

This is quite unintuitive for users, and leads to issues like #288  where a natural inheritance hierarchy with a partial'd base gets blocked.

### Solution

We can detect when a child is both inheriting `_partial_=True` and is using zen_processing, and we can overwrite `_partial_=False` and `_zen_partial=True` so that only `zen_processing` is responsible for performing the partial instantiation.

Thus we would have:

```python
P = builds(int, zen_partial=True)  # has _partial_ = True
builds(int, zen_partial=true_or_false, builds_bases=(P,), zen_meta=dict(a=1))  # has _partial_ = False, zen_partial=True
```

### Some sharp edges:

If we end up with configs like:

```
_zen_target: builtins.int
_partial_: false
_zen_partial: true
```

users may coerce their builds into being partial-builds by checking for / setting `_partial_=True`. This would be a mistake if `_zen_partial=True` -- it can lead to the "double partial" scenario laid out above. There are a few things that we might do to reduce the risks of this:

- Provide a public `is_partial_builds` function that will check both `_partial_` and `zen_partial` for users. This could either warn or raise in the case when both are set to True
- Provide a `validate_partial_builds` function that will raise when it sees bad combinations of `_partial_` and zen-processing

rsokl:
There are restrictions on inheritance patterns permitted by `builds`. Specifically, one could not inherit from a `PartialBuilds`-type config without explicitly "opting-in" and specifying `zen_partial=True` on the child. And even in this latter case, one could not include zen-processing features (such as `zen_meta`) in the child config.

See #289 for a detailed discussion of these restrictions.

This PR lifts all of these restrictions!

**Before:**

The default value for `zen_partial` was `False`. There was no way for a user to inherit from a partial-builds config and produce a builds-config

```python
parent_with_partial = builds(int, zen_partial=True)

builds(int)  # Type[Builds[Type[int]]]
builds(int, builds_bases=(parent_with_partial,)) # Not allowed
builds(int, zen_partial=False, builds_bases=(parent_with_partial,)) # Not allowed
builds(int, zen_partial=True, builds_bases=(parent_with_partial,)) # Type[PartialBuilds[Type[int]]]
builds(int, zen_partial=True, builds_bases=(parent_with_partial ,), zen_meta=dict(a=1))  # Not allowed (child uses zen-processing features)
```

**After:**

The default value for `zen_partial` is now `None`. Specifying `zen_partial` explicitly will determine whether the resulting config is a partial-builds, otherwise it is inherited or defaults to `False`

```python
parent_with_partial = builds(int, zen_partial=True)

builds(int)  # Type[Builds[Type[int]]]
builds(int, builds_bases=(parent_with_partial,)) # Type[PartialBuilds[Type[int]]]
builds(int, zen_partial=False, builds_bases=(parent_with_partial,))  # Type[Builds[Type[int]]]
builds(int, zen_partial=True, builds_bases=(parent_with_partial,)) # Type[PartialBuilds[Type[int]]]
builds(int, zen_partial=True, builds_bases=(parent_with_partial,), zen_meta=dict(a=1))  # Type[PartialBuilds[Type[int]]]
```

It achieves this by changing the default value of `builds(..., zen_partial)` to be `None` instead of `False`. 
- `None` indicates that the config will not entail partial instantiation unless that behavior is inherited
- `False` indicates will not entail partial instantiation under any circumstances

`make_config`  will now also raise when the following patterns are detected in a config:
- `_zen_target` is specified but `_target_` does not point to `hydra_zen.funcs.zen_processing`. This indicates a multiple inheritance pattern that "clobbers" the zen-processing configuration pattern
- `_partial_=True` is specified in conjunction with zen-processing patterns

## To-Do:
- [x] Add error handling in `make_config` when inheriting `_partial_=True` and `_zen_partial=True` fields simultaneously
- [x] Provide public function for checking `is_partial` and/or `validate_partial` so that users can inspect whether or not a config is a partial config without needing insights into the two different protocols for partial configs.

Closes #289 
Closes #288


rsokl:
@jgbos would you mind reviewing this / working off this branch during your day-to-day (if you aren't already) to confirm that you don't hit bugs? I still have some lingering to-dos, but they are relatively minor compared to the rest of this PR

jgbos:
@rsokl fyi, I haven't been running into any issues with the branch all week (I haven't done a lot though).

rsokl:
@jgbos great! I plan to try to tackle the outstanding To-Dos when possible and then merge this

src/hydra_zen/structured_configs/_type_guards.py
```diff
@@ -76,6 +127,55 @@ def uses_zen_processing(x: Any) -> TypeGuard[Builds[Any]]:
 
 
 def is_partial_builds(x: Any) -> TypeGuard[PartialBuilds[Any]]:
+    """
+    Returns `True` if the input is a targeted structured config that uses partial
```
rsokl, line 131:
```suggestion
    Returns `True` if the input is a targeted structured config that entails partial
```

PR #303, opened 2022-09-01
Adds auto-config support for `pydantic.Field`

rsokl:
Given:

```python
from pydantic.dataclasses import dataclass as pyd_dataclass
from pydantic import Field

@pyd_dataclass
class A:
    x: int = Field(default=2, gt=0, lt=3)
```

**Before**:
```python
>>> from hydra_zen import builds
>>> builds(A, populate_full_signature=True)  # presence of `pydantic.Field` causes error
HydraZenUnsupportedPrimitiveError: Building: A ..
 The configured value default=2 gt=0 lt=3 extra={}, for field `x`, is not supported by Hydra -- serializing or instantiating this config would ultimately result in an error.

Consider using `hydra_zen.builds(<class 'pydantic.fields.FieldInfo'>, ...)` create a config for this particular value.
```

**After**
```python
>>> from hydra_zen import builds, just, instantiate
>>> Conf = builds(A, populate_full_signature=True)
>>> instantiate(Conf)
A(x=2)
>>> just(A.x)
2
```

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -148,7 +148,7 @@ with Hydra. For example, a :py:class:`complex` value can be specified directly v
    _args_:
    - 3
 
-hydra-zen provides specialized support for values of the following types:
+hydra-zen provides specialized auto-config support for values of the following types:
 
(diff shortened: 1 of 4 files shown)


rsokl:
close?

PR #313, opened 2022-10-05
Allow dictionary for overrides in `launch`

jgbos:
Currently `launch` requires overrides to be a list of strings of type `param=value`. 

This PR allows the user to provide a dictionary as an override: `{param: val, ...}`. Where `param` is a string and `val` can be one of `int`, `float`, `str`, and `bool` along with with two special types:
  - `hydra_list`: A list type supported by Hydra, e.g., `hydra_list([1,2,3]) -> "[1,2,3]"`
  - `multirun`: A list that will be converted into a Hydra multirun value, e.g., `multirun([1,2,3]) -> "1,2,3"`

## Override Types

```python

override_primitive_examples = {
    "str": "\(1,2\)",   # --> "\(1,2\)" (no change)
    "NULL": None,   # --> "null"
    "int": 10,  #  --> "10"
    "float" 3.14,  # --> "3.14"
    "bool": True,    # --> "true"
    "list": hydra_list([1, 2, 3]),  # --> "[1, 2, 3]"
    "multirun_list": multirun([1, 2, 3])  # --> "1,2,3"
}
```

## Example of `multirun`

The use of the `multirun` type is what really makes this PR special.  With the current `overrides` structure a user must create the string that contains the list of values to multirun overview.  

```python
import random
from hydra_zen import launch, instantiate, make_config

values_for_experiment = [random.uniform(0, 1) for i in range(10)]
ov = []

jobs = launch(
    make_config(),
    instantiate,
    overrides=[
         "+param=" + ",".join([str(i) for i in values_for_experiment])
    ],
    multirun=True
)
```

This PR allows the user to provide the multirun value without having to create the string:

```python
import random
from hydra_zen import launch, instantiate, make_config, multirun

values_for_experiment = [random.uniform(0, 1) for i in range(10)]

jobs = launch(
    make_config(),
    instantiate,
    overrides={
         "+param": multirun(values_for_experiment)
    },
    multirun=True
)
```

TODO:
  - [ ] Update this PR description
  - [ ] Add `null` support in supported types
  - [ ] Test configs are the same for list of strings and dictionary
  - [ ] Update docstrings
  - [ ] Update docs


src/hydra_zen/_launch.py
```diff
@@ -1,8 +1,22 @@
...
+    List,
+    Mapping,
+    Optional,
+    Tuple,
+    Type,
+    TypeAlias,
+    TypeVar,
+    Union,
+    cast,
+)
```
rsokl, line 19:
```suggestion
    TypeVar,
    Union,
    cast,
)
from typing_extensions import  TypeAlias
```

src/hydra_zen/_launch.py
```diff
@@ -232,6 +319,15 @@ def launch(
     else:
         config_name = _store_config(config, config_name)
 
+    # allow user to provide a dictionary of override values
+    # instead of just a list of strings
+    overrides = overrides if overrides is not None else []
+    if isinstance(overrides, Dict):
```
rsokl, line 325:
```suggestion
    if isinstance(overrides, dict):
```

docs/source/explanation/launch.rst
```diff
@@ -0,0 +1,138 @@
...
+   def task_fn(foo):
+      print(foo)
+
+
+   # run two Hydra jobs within one script
+   job = launch(TestConfig, task_fn, overrides=["+foo=1"])
+   # outputs:
+   # 1
+
+   mulritun_job = launch(TestConfig, task_fn, overrides=["+foo=1,2,3"], multirun=True)
```
rsokl, line 83:
```suggestion
   multirun_job = launch(TestConfig, task_fn, overrides=["+foo=1,2,3"], multirun=True)
```

docs/source/explanation/launch.rst
```diff
@@ -0,0 +1,138 @@
...
+   # outputs:
+   # [2023-01-11 10:56:02,448][HYDRA] Launching 3 jobs locally
+   # [2023-01-11 10:56:02,448][HYDRA]        #0 : +foo=1
+   # 1
+   # [2023-01-11 10:56:02,537][HYDRA]        #1 : +foo=2
+   # 2
+   # [2023-01-11 10:56:02,626][HYDRA]        #2 : +foo=3
+
+   # define a standard Hydra list as a single parameter using `hydra_zen.hydra_list`
+   mulritun_job = launch(TestConfig, task_fn, overrides={"+foo": hydra_list([1, 2, 3])})
```
rsokl, line 128:
```suggestion
   multirun_job = launch(TestConfig, task_fn, overrides={"+foo": hydra_list([1, 2, 3])})
```

src/hydra_zen/_launch.py
```diff
@@ -111,6 +217,10 @@ def launch(
     with_log_configuration : bool (default: True)
         If ``True``, enables the configuration of the logging subsystem from the loaded config.
 
+    **override_kwargs : OverrideValues
+        Keyword arguments to override existing configuration values.  Note, this only works when the configuration value,
```
rsokl, line 221: "when the configuration value" .. what? (seems like an incomplete sentence)

rsokl (changes requested):
Looks good at a first pass. I'll do a more thorough review this evening

rsokl (approved):
Looks good!

PR #384, opened 2023-01-10
Improve zen pickle-compat and support for `hydra_main(config_path)`

### LINKED ISSUE #381: `config_path` broken for `hydra_main`
Hello, it seems like `config_path` is not functional when used with the new `@zen` + `hydra_main` pattern introduced in 0.9.0.

Here's how I repro:

File `/home/petosa/repro/config.yaml`
```
db:
  host: local
  port: 3307
```
File `/home/petosa/repro/zen.py`
```
from hydra_zen import zen

@zen
def main(zen_cfg):
    print(zen_cfg)

if __name__ == "__main__":
    main.hydra_main(config_name="conf", verion_base="1.1", config_path="/home/petosa/repro")
```

File `/home/petosa/repro/hy.py`
```
import hydra

@hydra.main(config_name="conf", verion_base="1.1", config_path="/home/petosa/repro")
def main(cfg):
    print(cfg)

if __name__ == "__main__":
    main()

```

Running `hy.py` successfully loads and prints the config.

Running `zen.py` gives exception 
`Primary config module 'hydra_zen.wrapper..home.petosa.repro' not found.`

rsokl:
- Makes `Zen` instances pickle-compatible
- Fixes `Zen.hydra_main(config_path)` functionality for Hydra 1.3.0+ (Closes #381) 

By default `hydra.main(config_path)` treats `config_path` as being relative to the directory that houses the file in which the task function is defined. Hydra has some relatively complicated [logic for determining this directory](https://github.com/facebookresearch/hydra/blob/1e350d17c7aa75e14839474f9e4028ad0fb4a581/hydra/_internal/utils.py#L37-L88).

Because `zen(task).hydra_main(...)` sends a zen-wrapped task function to `hydra.main`, the location of `task` cannot be determined. I did not appreciate what was going on under the hood here until #381 was opened. Fortunately, Hydra 1.3.0 added support for wrapped task functions (i.e. it will follow a `__wrapped__` attribute to find the task function.

Thus, in the case that `config_path` is a string and Hydra 1.3.0+ is installed, `zen` will now produce a temporary wrapped function that points to the original task function. The reason why `zen` doesn't *always* do this is that the resulting wrapped function is not pickle-compatible, whereas the zen-instance is. It is possible in the future that pickle-compatible task functions could be useful for some launchers.

Unfortunately, I was not able to find a way to make `config_path=<str>` work for Hydra 1.2.0 or earlier. Thus we emit a warning to users.

Given:
```
dir/
  |- bug_main.py
  |- inner/
       |- config.yaml
```
config.yaml
```
db:
  host: local
  port: 3307
```
```python
# contents of bug_main.py

from hydra_zen import zen

cwd = Path.cwd()


@zen
def main(zen_cfg):
    print(zen_cfg)


if __name__ == "__main__":
    main.hydra_main(
        config_name="config", version_base="1.1", config_path="."
    )
```

**Before**:

```console
$ python bug_main.py --config-path=inner
Primary config module 'hydra_zen.wrapper.inner' not found.
Check that it's correct and contains an __init__.py file

Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.
```

**After**
(Hydra 1.3.0+ only)

```console
$ python bug_main.py --config-path=inner
{'db': {'host': 'local', 'port': 3307}}
```


rsokl:
Note to self: add pickle compat test

jgbos (approved):

Jasha10:
> is there anything hazardous with the [solution I landed on](https://github.com/mit-ll-responsible-ai/hydra-zen/blob/3a0ede6506fc70f35559cd30e8e4f81d890d02ca/src/hydra_zen/wrapper/_implementations.py#L405-L441)?

I.e. using a wrapper? It looks good to me!

> Any ideas about making this work for Hydra < 1.3.0?

Since the `hydra/_internal/utils.py:detect_calling_file_or_module_from_task_function` function that you mentioned above inspects `task_function.__module__`, a dirty hack would be to set `task_function.__module__` manually inside of the `Zen.hydra_main` method:
```python
...
        if (config_path is _UNSPECIFIED_ and HYDRA_VERSION < Version(1, 2, 0)) or (
            isinstance(config_path, str) and HYDRA_VERSION < Version(1, 3, 0)
        ):  # pragma: no cover
            target = self
            target.__module__ = self.func.__module__
        elif Version(1, 3, 0) <= HYDRA_VERSION and isinstance(config_path, str):
            ...
            target = wrapper
        else:
            target = self
...
```
To maintain some sense of decorum, you could even use a `try`/`except` block to reset the `self.__module__` attribute after the call `hydra.main(**kw)(target)()` has finished running.

src/hydra_zen/wrapper/_implementations.py
```diff
@@ -394,6 +405,32 @@ def hydra_main(
...
+        # in which the task function is defined. Unfortunately, it is only able to
+        # follow wrappers starting in Hydra 1.3.0. Thus `Zen.hydra_main` cannot
+        # handle string config_path entries until Hydra 1.3.0
+        if (config_path is _UNSPECIFIED_ and HYDRA_VERSION < Version(1, 2, 0)) or (
+            isinstance(config_path, str) and HYDRA_VERSION < Version(1, 3, 0)
+        ):  # pragma: no cover
+            warnings.warn(
+                "Specifying config_path via hydra_zen.zen(...).hydra_main "
+                "is only supported for Hydra 1.3.0+"
+            )
```
Jasha10, line 418: The warning message would only seem to be relevant if `config_path` has been specified. Checking whether `config_path is _UNSPECIFIED_` in this `if` statement seems strange.

rsokl:
> I.e. using a wrapper? It looks good to me!

Great! 🚀 

> a dirty hack would be to set task_function.__module__ manually inside of the Zen.hydra_main method

This is an interesting idea. I tried this and it seems that when the module becomes `__main__`, as in many cases, Hydra looks to the call stack, where ends up finding the hydra-zen internals :/ . 

I haven't been able to find a complete backwards-compat w/ Hydra solution, so I think I will just punt and promise functionality for Hydra 1.3.0+. Fortunately the upgrade from 1.2 to 1.3 should be quite benign for most people.

Thanks for the help @Jasha10 !

src/hydra_zen/wrapper/_implementations.py
rsokl: I did this because if `version_base` is Hydra 1.1 and `config_path` is `_UNSPECIFIED_` then Hydra defaults `config_path` to be `"."`

But I realize that I need to have additional logic here to handle the case where the `version_base` via `zen` is specified

Issue #394, opened 2023-01-18
hydra-zen not calling os.chdir


DubiousCactus:
Hi,

I'm already quite familiar with hydra and just started using hydra-zen for my new project. I am struggling a little and I think the examples lack consistency (why do I have to do things differently in the [pytorch example](https://mit-ll-responsible-ai.github.io/hydra-zen/how_to/pytorch_lightning.html) than in the tutorials?).
Anyway, when I save files and create directories, they appear at the root of my project and not in the `outputs/` tree as they would when using hydra.
Here is the entry point of my program:
```python

def launch_experiment(exp: BaseExperiementConfig):
    hydra_zen.save_as_yaml(exp, "experiment.yaml")
    "============ CUDA ============"
    model: torch.nn.Module = to_cuda(exp.model)  # type: ignore

    "============ Weights & Biases ============"
    if project_conf.USE_WANDB:
        wandb.init(
            project=project_conf.PROJECT_NAME,
            # config=store(exp),
            # reinit=True, What does this do?
        )

    " ============ Partials instantiation ============ "
    opt = exp.opt(model.parameters())
    train_dataset = exp.train_dataset(tiny=exp.tiny_dataset)
    val_dataset = exp.val_dataset(tiny=exp.tiny_dataset)
    train_batch_sampler = exp.train_batch_sampler(train_dataset)
    val_batch_sampler = exp.val_batch_sampler(val_dataset)

    " ============ Reproducibility of data loaders ============ "
    g = None
    if project_conf.REPRODUCIBLE:
        g = torch.Generator()
        g.manual_seed(0)

    train_loader = exp.train_loader(
        train_dataset, batch_sampler=train_batch_sampler, generator=g
    )
    val_loader = exp.val_loader(
        val_dataset, batch_sampler=val_batch_sampler, generator=g
    )

    " ============ Training ============ "
    BaseTrainer(
        model=model,
        opt=opt,
        train_loader=train_loader,
        val_loader=val_loader,
    ).train(
        epochs=exp.epochs,
        val_every=exp.val_every,
        visualize_every=exp.visualize_every,
    )


if __name__ == "__main__":
    " ============ Hydra-Zen ============ "
    store.add_to_hydra_store()
    zen(
        launch_experiment,
        pre_call=hydra_zen.zen(
            lambda exp: seed_everything(exp.seed) if project_conf.REPRODUCIBLE else lambda: None
        ),
    ).hydra_main(
[...]

PR #398, opened 2023-01-19
Add Examples to Docs

jgbos:
This PR updates the docs based on discussion in:

- [x] #389 
- [x] #394 
- [x] #397


rsokl:
@jgbos assuming that CI passes, this should be good to go. I made a good number of changes, so it would be great if you could build the docs locally and give the new sections a read through /validate that they run.

Thanks so much for putting these together! I think that they are great additions to the docs!

rsokl (approved):

jgbos:
Wow, thanks for the updates and making the examples clearer.  I expected to do a few more updates but this all looks really great.  Lets merge and make this live!

PR #401, opened 2023-01-23
Infer name/group for storing HydraConf instances/subclasses

### LINKED ISSUE #395: Specialized support for `HydraConf` in `ZenStore`?
Inspired by #394 , it might be nice for `store`/`ZenStore` to have specialized support for instances/subclasses of `HydraConf` so that

```python
store(HydraConf(job=JobConf(chdir=True)))
```

is equivalent to the long-form

```python
store(HydraConf(job=JobConf(chdir=True)), name="config", group="hydra", provider="hydra_zen")
```

i.e. we auto-populate the name/group/provider to be the obvious entries if the user doesn't specify it.

Furthermore, we can special case `store.add_to_hydra_store()` to automatically overwrite the default Hydra entry (but only if it is the entry provided by Hydra itself) since this is the obvious desired behavior if the user is storing their own `HydraConf`.

This would almost certainly be the only special case we would want to specify for `store`. There are some tradeoffs here between convenience and explicitness, and perhaps this leans towards making things a bit too magical for users. It is worth discussing the pros and cons.

rsokl:
Closes #395

### DIFF
--- a/src/hydra_zen/wrapper/_implementations.py
+++ b/src/hydra_zen/wrapper/_implementations.py
@@ -914,6 +914,28 @@ class ZenStore:
            deferred_hydra_store=True,
        )
 
+    Notes
+    -----
+    Special support is provided for overriding Hydra's configuration; the name and
+    group of the store entry is inferred to be 'config' and 'hydra', respectively,
+    when an instance/subclass of `HydraConf` is being stored. E.g., specifying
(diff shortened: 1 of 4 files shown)


jgbos:
Looks great to me!

jgbos (approved):

PR #448, opened 2023-04-08
Drop support for hydra-core < 1.2.0. Improve `zen` to return dataclasses and stdlib containers over omegaconf objects

### LINKED ISSUE #442: ConfigDicts aren't converted to dataclasses when specified in a list
The conversion back to dataclasses doesn't seem to work, when they are specified in a list.
Here is a minimal example.
```python
from collections.abc import Sequence
from hydra_zen import store, zen, builds
from hydra_zen.typing import ZenConvert

@dataclass
class B:
    x: int

def app(dc: Sequence[B]) -> None:
    assert all(isinstance(b, B) for b in dc)

Config = builds(
    app,
    populate_full_signature=True,
    dc=[B(x=0)],
    zen_convert=ZenConvert(dataclass=True),
    hydra_convert='all'
)
store(Config, name="config")

if __name__ == "__main__":
    store.add_to_hydra_store()
    zen(app).hydra_main(config_path=None, config_name="config", version_base="1.3")
```
It fails with an assertion error. The entry in `dc` is a `DictConfig`.

Thanks!

### LINKED ISSUE #450: Drop support for omegaconf < 2.2.1 and hydra-core < 1.2.0
Will enable #448  and will remove a whole lot of hot-patch code

rsokl:
Update `zen` to return dataclasses and stdlib containers over omegaconf objects

Closes #442
Closes #450 

Requires min hydra-core -> 1.2.0; min omegaconf -> 2.2.1, which I was thinking of doing anyway -- would simplify lots of logic due to hot patching old omegaconf issues

## TODO:
- [ ] write PR summary
- [ ] add changelog write up
- [ ] think about potential issues with `zen`
- [ ] document new `zen` method


rsokl:
Well that was satisfying

image

PR #454, opened 2023-04-15
Add `flat_target` setting to `builds`

### LINKED ISSUE #453: Nested Partial Builds not Converted when Using Stores and Hydrated_dataclass with Zen_partial
When using stores and hydrated_dataclass together with zen_partial, it seems that the _target_ field gets assigned the Config instead of the actual target. This results in functools.partial[<class '__main__.NestedModelConf'>] instead of functools.partial[<class '__main__.NestedModel'>]. However, NestedModel is also a partial, which results in a double partial, and the need for manually initializing the config in the outer model.

```python
from dataclasses import dataclass, field
from hydra_zen import (
    MISSING,
    hydrated_dataclass,
    instantiate,
    zen,
    store,
)
from hydra_zen.typing import Partial
from omegaconf import DictConfig, OmegaConf
from torch import nn


class NestedModel(nn.Module):
    def __init__(self, **kwargs):
        super().__init__()
        params = NestedModelConf(**kwargs)
        self.linear = nn.Linear(params.in_features, 10)

    def forward(self, x):
        return self.linear(x)


class Model(nn.Module):
    def __init__(self, **kwargs):
        super().__init__()
        params = ModelConf(**kwargs)
        self.linear = nn.Linear(params.in_features, params.out_features)
        self.nested = params.nested(params.out_features)

    def forward(self, x):
        return self.linear(x)


@store(name="x", group="model/layer")
@hydrated_dataclass(NestedModel, zen_partial=True)
class NestedModelConf:
    in_features: int = MISSING


@hydrated_dataclass(nn.Module)
class BaseModelConf:
    pass


@store(name="mod", group="model", in_features=10, out_features=20)
@hydrated_dataclass(Model)
class ModelConf(BaseModelConf):
    in_features: int = 1
    out_features: int = 1
    nested: Partial[NestedModel] = MISSING


def task(zen_cfg: DictConfig):
    cfg: Exp = instantiate(zen_cfg)
    print(cfg.x + cfg.y)
    model = instantiate(cfg.model)


@dataclass
class Exp:
    x: int
    y: int
    model: BaseModelConf


cs = ConfigStore.instance()
cs.store(name="my_app", node=Exp)

@store(name=lambda cfg: cfg.__name__, group="exp", package="_global_")
@dataclass
class OverrideExp(Exp):
    x: int = 52
    y: int = 30
    model: BaseModelConf = MISSING
    defaults: list = field(
        default_factory=lambda: [
            "_self_",
            {
                "/model": "mod",
            },
            {
                "/model/layer@model.nested": "x",
            },
        ]
    )


if __name__ == "__main__":
    store.add_to_hydra_store()
    zen(task).hydra_main(config_name="my_app", config_path=None, version_base="1.3")
```

The line `self.nested = params.nested(params.out_features)` is not valid, as it simply returns NestedModelConf(in_features=20, _target_='__main__.NestedModel', _partial_=True). In order to get the actual target, I have to execute instantiate(self.nested). This will give me functools.partial(<class '__main__.NestedModel'>, in_features=20), which again I have to call.

I hope this makes sense and that I used the library correctly.

To run this code, I executed: python main.py +exp=OverrideExp

For completion's sake, this is the produced job YAML:

```yaml
x: 52
'y': 30
model:
  _target_: __main__.ModelConf
  in_features: 10
  out_features: 20
  nested:
    in_features: ???
    _target_: __main__.NestedModelConf
    _partial_: true
_target_: __main__.OverrideExp
```
For now I resolved the problem by applying this function to the task config: 

```python
def fix_partials(cfg: DictConfig):
    for k, v in cfg.items():
        if isinstance(v, DictConfig):
            if "_partial_" in v:
                v._target_ = v._target_.replace("Conf", "")
                return cfg
            cfg[k] = fix_partials(v)

    return cfg 
```

However, this is a rather hacky solution and requires exact naming.

A few additional notes:
```python
@dataclass
class Exp:
    x: int
    y: int
    model: BaseModelConf


cs = ConfigStore.instance()
cs.store(name="my_app", node=Exp)
```
is required, because If I use the builtin hydra_zen store on either the task or this base experiment, OmegaConf, cannot merge any subclass Experiment with this one. Apparently `Builds_Exp` and `Builds_OverrideExp` are not of the same base type.

Thank you for your help.

rsokl:
(git history is all messed up because I did a stupid rebase)

Fixes / Closes #453

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,6 +112,9 @@ exclude = [
        "**/__pycache__",
        "src/hydra_zen/_version.py",
        "**/third_party",
+       "**/.hypothesis",
+       "**/docs",
+       "**/.tox",
 ]
 reportUnnecessaryTypeIgnoreComment = true
 reportUnnecessaryIsInstance = false
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -57,6 +57,7 @@ from hydra_zen.typing import (
 )
 from hydra_zen.typing._implementations import (
     AllConvert,
+    AnyBuilds,
     BuildsWithSig,
     DataClass_,
     DefaultsList,
@@ -96,7 +97,7 @@ R = TypeVar("R")
 Field_Entry: TypeAlias = Tuple[str, type, Field[Any]]
 
 # default zen_convert settings for `builds` and `hydrated_dataclass`
-_BUILDS_CONVERT_SETTINGS = AllConvert(dataclass=True)
+_BUILDS_CONVERT_SETTINGS = AllConvert(dataclass=True, flat_target=True)
 
 # stores type -> value-conversion-fn
 # for types with specialized support from hydra-zen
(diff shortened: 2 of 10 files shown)


rsokl:
#455  because I am so scared of losing this to a bad rebase

PR #497, opened 2023-06-30
Update `builds` signature inspection for python 3.11.4 change

rsokl:
https://github.com/python/cpython/issues/105080 updates the behavior of `inspect.signature` to walk through a class' mro in search of its signature, rather than just defer to the most recent definition of `__new__`.

This PR updates the signature inspection logic of `builds` to match this new behavior of `inspect.signature` across all versions of Python. I.e. this effectively serves as a backport.

Here is an example that exhibits this change in behavior:

```python
class A:
    def __new__(cls, a: int):
        return object.__new__(cls)

class B(A):
    def __init__(self, b: float):  ...

class C(B):
    ...
```

Before:

The signature of `A.__new__` was preferred over that of `B.__init__` even though the latter is closer to `C` in the inheritance hierarchy.

```python
>>> import inspect
>>> from hydra_zen import inspect
>>> inspect.signature(builds(c, populate_full_signature=True))
<Signature (a: int) -> None>
```

After:

C's signature is that of `B.__init__`

```python
>>> import inspect
>>> from hydra_zen import inspect
>>> inspect.signature(builds(c, populate_full_signature=True))
<Signature (b: float) -> None>
```

### DIFF
--- a/docs/source/changes.rst
+++ b/docs/source/changes.rst
@@ -51,7 +51,7 @@ hydra-zen now uses the `trusted publishers <https://blog.pypi.org/posts/2023-04-
 Improvements
 ------------
 - :func:`~hydra_zen.builds` now has a transitive property that enables iterative build patterns. See :pull:`455`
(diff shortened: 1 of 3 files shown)


rsokl:
Merging with failures caused by a separate compatibility issue with the latest numpy

PR #499, opened 2023-06-30
Bump pyright to 1.1.317

rsokl:
I expect one failure associated with the overloads for `zen` - I believe this is a pyright bug: https://github.com/microsoft/pyright/issues/5416.

There are many inline "iterative builds" patterns that pyright mishandles currently. This was already reported in https://github.com/microsoft/pyright/issues/5166, which I will reopen.


rsokl:
There is another pyright inconsistency with inline expression:

```python
a = zen(lambda x: 1)
zen(lambda: None, pre_call=a)  # okay
zen(lambda: None, pre_call=zen(lambda x: 1))  # not okay
```

PR #557, opened 2023-10-25
Add `zen_exclude` option to `builds.

rsokl:
```python
def f(x=1, y=2, _z=3): ...

builds(f, populate_full_signature=True, zen_exclude=['y'])  # exclude `y` from config
builds(f, populate_full_signature=True, zen_exclude=lambda x: x.startwith("_")  # exclude any param starting with "_"

```


rsokl:
Closed in favor of #558

PR #565, opened 2023-11-01
add `hydra_zen.note_static_method`

rsokl:
This PR adds the ability to pass a static method to `builds` by first "noting" the static method via `hydra_zen.note_static_method` or `hydra_zen.BuildsFn.note_static_method`.

To be able to pass a static method to `builds` do:

```python
from hydra_zen import note_static_method, builds

class Foo:
    @staticmethod
    def f():
        return "BARK"

Conf = builds(note_static_method(Foo.f))
```

## Explanation

Consider the class:

```python
class Foo:
    @staticmethod
    def f():
        return "BARK"
```

Suppose that we want to make a config that will instantiate `Foo.f`. If we pass this static method to `builds` there is an issue:

```python
>>> from hydra_zen import builds
>>> builds(Foo.f)._target_
'__main__.f'
```

The correct `_target_` is `"__main__.Foo.f"`. `builds` is unable to find the correct import path for this static method because there is no way to distinguish a static method from a function once that static method has been attached to a class.

We can call

```python
from hydra_zen import note_static_method

note_static_method(Foo.f)
```

and this will register the appropriate import path for `Foo.f` in a registry associated with `BuildsFn`. And now `builds` can resolve the correct path:

```python
>>> builds(Foo.f)._target_
'__main__.Foo.f'
```

## Usage patterns

`note_static_method` need only be called on a given static method once. All of the following patterns are valid

```python
# "noting" prior to passing to `builds`
note_static_method(Foo.f)
builds(Foo.f)
```

```python
# noting inline within `builds`
builds(note_static_method(Foo.f))
```

For Python 3.10+ you can use `note_static_method` as a decorator:

```python
class Foo:
    @note_static_method
    @staticmethod
    def f():
        return "BARK"

builds(Foo.f)
```
(It is also possible to combine `note_static_method` and `staticmethod` into a single decorator[^scary])

## Implementation Details
`note_static_method` registers a tuple of strings to a set held by `BuildsFn`. If you want to isolate the registry that is populated and used by config-creation functions, do:

```python
from hydra_zen import BuildsFn

class MyBuilds(BuildsFn):
    _registered_static_methods = set()

builds = MyBuilds.builds
note_static_method = MyBuilds.note_static_method
```

[^scary]: 
    ```python
    # Merging `note_static_method` and `staticmethod` into one decorator.
    # Use at your own risk!
    from typing import TypeVar

    T = TypeVar("T")

    def st(__x: T) -> T:
        def merged(f):
            return  note_static_method(staticmethod(f))  # type: ignore
        return merged  # type: ignore

    new_static_method = st(staticmethod)

    class BB:
        @new_static_method
        def foo():
            print("meow")

    # `builds(BB.foo)` will "automatically" work now
    ```

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -38,6 +38,7 @@ Creating Configs
    builds
    just
    hydrated_dataclass
+   note_static_method
 
 
 Storing Configs
--- a/docs/source/changes.rst
+++ b/docs/source/changes.rst
@@ -67,6 +67,7 @@ Improvements
 - Adds formal support for Python 3.12. See :pull:`555`
 - :func:`~hydra_zen.BuildsFn` was introduced to permit customizable auto-config and type-refinement support in config-creation functions. See :pull:`553`.
 - :func:`~hydra_zen.builds` and :func:`~hydra_zen.make_custom_builds_fn` now accept a `zen_exclude` field for excluding parameters from auto-population, either by name or by pattern. See :pull:`558`.
+- :func:`~hydra_zen.note_static_method` and :func:`~hydra_zen.BuildsFn.note_static_method` have been added to enable support for passing a static method to :func:`~hydra_zen.builds`. See :pull:`565`.
 
 
 .. _v0.11.0:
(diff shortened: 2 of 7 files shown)


rsokl:
I just realized that I can get this to work "automatically":

```python
        if not _utils.is_classmethod(target):
            if (
                inspect.isfunction(target)
                and isinstance(qualname, str)
                and "." in  qualname
            ):
                # this looks like it is a staticmethod
                return f"{module}.{qualname}"
            return f"{module}.{name}"
```

welp.. wish I had realized that before writing this whole PR

rsokl:
But the good news is that we get static method support with no additional steps!

PR #566, opened 2023-11-02
Add automatic config-creation support for staticmethods

rsokl:
hydra-zen's config creation functions now have special logic for determining the import-path of a static method. E.g.

Given:

```python
class A:
    @staticmethod
    def f(): return "woo"
```

Before:

```python
>>> from hydra_zen import builds, instantiate
>>> instantiate(builds(A.f))
--------------------------------------------
InstantiationException: Error locating target '__main__.f', set env var HYDRA_FULL_ERROR=1 to see chained exception.
```

After:

```python
>>> from hydra_zen import builds, just, instantiate
>>> instantiate(builds(A.f))                 
woo
>>> instantiate(just(A.f))
<function __main__.A.f()>
```

## Implementation details
It is not possible to distinguish a staticmethod that has already been bound to a class versus a function. That being said, we can inspect the `__qualname__` of a function to see if it is distinct from the function's `__name__` and if it looks something like `SomeClass.some_func`. 

There may be some edge cases / false positives, but the net result of supporting static methods is very much worth it. I am also more willing to include this logic now that people can customize `BuildsFn._get_obj_path` in case, somehow, this change really breaks their application.

### DIFF
--- a/docs/source/changes.rst
+++ b/docs/source/changes.rst
@@ -64,9 +64,10 @@ For more details and examples, see :pull:`553`.
 
 Improvements
 ------------
-- Adds formal support for Python 3.12. See :pull:`555`
 - :func:`~hydra_zen.BuildsFn` was introduced to permit customizable auto-config and type-refinement support in config-creation functions. See :pull:`553`.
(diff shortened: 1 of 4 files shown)


jgbos:
Awesome how simple that is

PR #571, opened 2023-11-08
Make ZenStore.update noop for identical stores

rsokl:


rsokl:
- A basic overview of Hydra and hydra-zen
  - What do they do and how do they differ?
  - Hydra: compose configs, generate CLI, expose grammar, launch & log
  - hydra-zen: make & store configs. Pass configs to functions that aren't designed to accept configs.

- What is a config?
  - yaml
  - dataclass w/ recursively yaml-serializable fields
  - omegaconf container
  - Explicit is better than implicit, unless it means repetition
    - The 

- Anatomy of a Hydra(zen) app
  - Design a task function
  - Make and store configs
  - Launch

PR #638, opened 2024-01-27
make flat-target path point to correct target with zen-processing

rsokl:
Potentially addresses https://github.com/mit-ll-responsible-ai/hydra-zen/discussions/637

zen-processing features can be used with the flat-target feature via inheritance:

```python
from hydra_zen import builds

A = builds(dict, x=1, zen_meta={"META": 2})
B = builds(A, x="${META}", builds_bases=(A,))
```


src/hydra_zen/structured_configs/_implementations.py
```diff
@@ -3465,6 +3466,43 @@ def __post_init__(self, CBuildsFn: Type[BuildsFn[Any]]) -> None:  # pragma: no c
         del CBuildsFn
 
 
+def get_target_path(obj: Union[HasTarget, HasTargetInst]) -> Any:
+    """
+    Returns the target-object from a targeted config.
```
rsokl, line 3471:
```suggestion
    Returns the import-path from a targeted config.
```

PR #708, opened 2024-07-12
Update _get_obj_path to support ClassA.ClassB

### LINKED ISSUE #705: _target_ for inner classes is incorrect
Hello there, big fan of this package.

I believe that there is a bug with the `safe_name` function, that causes the `_target_` field of dynamic configs to be wrong if the target is an inner class:

```python
# project.configs.algorithm.__init__.py

import hydra_zen
import dataclasses

class MyAlgorithm:
    @dataclasses.dataclass
    class HParams:
        lr: float = 0.01



HparamsConfig = hydra_zen.builds(MyAlgorithm.HParams)
config = HparamsConfig()

# This should be true, but it isn't! The target is actually "project.configs.algorithm.HParams"!
assert config._target_ == "project.configs.algorithm.MyAlgorithm.HParams"
```

rsokl:
Closes #705

Given:

```python
from hydra_zen import builds, instantiate

class Outer:
    class Inner:
        pass

cfg = builds(Outer.Inner)
```

Before:

```python
>>> instantiate(cfg)
InstantiationException: Error locating target '__main__.Inner'
```

After:

```python
>>> instantiate(cfg)
Outer.Inner()
```

### DIFF
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -992,7 +992,9 @@ class BuildsFn(Generic[T]):
                 and "." in qualname
                 and all(x.isidentifier() for x in qualname.split("."))
             ):
-                # this looks like it is a staticmethod. E.g. qualname: SomeClass.func
+                # This looks like it is a staticmethod or a class defined within
+                # a class namespace. E.g. qualname: SomeClass.func or
+                # SomeClass.NestedClass
                 return f"{module}.{qualname}"
             return f"{module}.{name}"
         else:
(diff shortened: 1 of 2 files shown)


src/hydra_zen/structured_configs/_implementations.py
```diff
@@ -948,6 +948,20 @@ def _get_obj_path(cls, target: Any) -> str:
...
+
+            name = _utils.safe_name(obj, repr_allowed=repr_allowed)
+            qualname = obj.__qualname__
+            assert isinstance(qualname, str)
+
+            if name != qualname and qualname.endswith("." + name):
+                return qualname
+
+            return name
+
```
rsokl, line 964:
```suggestion
```

PR #731, opened 2024-09-20
Fixed builds auto-populating example doc

mirceamironenco:
I assume `f` is supposed to be `bar` in this example?


rsokl (approved):
Yep, thanks! 😄

PR #777, opened 2025-02-17
Refactor `_StoreCallSig` to inline form

sobolevn:
While working on https://github.com/python/mypy/pull/16715 I've noticed that `_StoreCallSig` uses class form, while having `__kw` field. This field will be mangled. This is not something typing users want. See https://github.com/python/cpython/issues/129567


rsokl:
Fixing formatting in #779

PR #870, opened 2026-05-18
Future compat omegaconf yaml loader

rsokl:


stanlrt:
Hey! Any idea when this is planned to get released? No time pressure, it's just to know when we can expect to remove the deps pins in pyproject files (some users use the pre-release flag so we have to be defensive)

rsokl:
Publishing a release candidate right now

stanlrt:
Wow amazing

Issue #574, opened 2023-11-08
pre-release CI job is flaky


rsokl:
I don't know what is going on [here](https://github.com/mit-ll-responsible-ai/hydra-zen/actions/runs/6801140098/job/18491313239?pr=573). It only occurs for pre-release, and re-running the job often fixes it. It occurs in `tests/test_launch/test_validation.py` and is caused by:

```
 hydra.errors.MissingConfigException: In 'hydra/config': Could not find 'hydra/sweeper/basic'
```

@jgbos  any ideas?

jgbos:
That is weird.  I'll try to look at it this afternoon.

rsokl:
@jgbos I think there is a real issue here, although I'm unsure what the culprit is. I'd like to get a fix in for launch before v0.12 is released. Any chance you'd be able to take a look this week?

rsokl:
This was really hard to repro -- still not sure the circumstances that cause this vs not -- but there is some broken statefulness in Hydra between runs which causes Hydra's store to misbehave with the basic sweeper. Fortunately there is a simple fix - running the following clears ~all global state from Hydra!

```python
from hydra.core.singleton import Singleton

# this should go in a `finally:` after each launch call
Singleton._instances.clear()
```


Because this clears all global state, Hydra will populate everything anew on subsequent runs.

PR #875, opened 2026-06-23
Fix just() conversion for container fields of dataclass types (#858)

martinez-hub:
## Summary

Fixes #858. `just` failed to convert a dataclass to a config when one of its fields is annotated with a **container of a dataclass type** (e.g. `dict[str, Inner]` or `list[Inner]`).

```python
import hydra_zen as hz

@dataclass
class Inner:
    x: int = 0

@dataclass
class Outer:
    outer: dict[str, Inner]

hz.to_yaml(hz.just(Outer(outer={"a": Inner(x=1)})))
# omegaconf.errors.ValidationError: Invalid type assigned:
# Builds_Inner is not a subclass of Inner. value: <class 'types.Builds_Inner'>
```

## Root cause

`just` recursively converts the dataclass *values* into `Builds_Inner` configs, but the field's annotation keeps the original element type (`dict[str, Inner]`). OmegaConf type-checks container elements before instantiation and rejects the `Builds_Inner` config since it isn't a subclass of `Inner`.

`_retain_type_info` already broadens a field's annotation to `Any` when its value *is* a structured config — but it only inspected the top-level value, not configs nested inside a container. A second gap: mutable container defaults are stored via `default_factory`, so the value reaching that check was `MISSING`.

## Fix

- `_value_contains_builds(value)` — detects a structured config nested (possibly deeply) inside a list/tuple/mapping.
- `_resolve_field_default(field)` — resolves a field's `default_factory` so the real container value can be inspected.
- `_retain_type_info` now broadens the annotation to `Any` when the value is a container holding a structured config.

Broadening to `Any` only loosens OmegaConf validation — it never tightens an annotation, so previously-working configs are unaffected.

## Tests

Added to `tests/test_dataclass_conversion.py`:
- `test_just_on_container_field_of_typed_dataclasses` — `dict[str, Inner]`, `list[Inner]`, and nested `dict[str, list[Inner]]`, each round-tripped through `to_yaml` → `instantiate`.
- Unit tests for both new helpers.

## Verification

- Full test suite passes locally (`test_launch`/`test_with_hydra_submitit`/`test_third_party` excluded — they run in CI).
- 100% coverage on the changed lines.
- `pyright` basic: 0 errors on the changed source.
- `ruff format --check`, `ruff check`, and `autoflake` all clean (ruff 0.14.1).
- Changelog entry added under a new `0.16.1` section.

### DIFF
--- a/docs/source/changes.rst
+++ b/docs/source/changes.rst
@@ -8,6 +8,14 @@ Changelog
 This is a record of all past hydra-zen releases and what went into them, in reverse
 chronological order. All previous releases should still be available on pip.
 
+.. _v0.16.1:
+
+-------------------
+0.16.1 - 2026-06-23
+-------------------
+
+- Fixes a bug where :func:`~hydra_zen.just` would fail to convert a dataclass to a config when one of its fields is annotated with a container of a dataclass type (e.g. ``dict[str, SomeDataclass]`` or ``list[SomeDataclass]``). Previously this raised an OmegaConf ``ValidationError`` during ``to_yaml``. See :issue:`858`.
+
 .. _v0.16.0:
 
 -------------------
--- a/src/hydra_zen/structured_configs/_implementations.py
+++ b/src/hydra_zen/structured_configs/_implementations.py
@@ -208,6 +208,34 @@ if not OMEGACONF_HANDLES_PATHLIB_LOCAL:  # pragma: no cover
     _omegaconf_utils.get_yaml_loader = _patched_yaml_loader
 
 
+def _value_contains_builds(value: Any) -> bool:
+    # Returns `True` if `value` is a list/tuple/mapping that contains
+    # (possibly nested) a structured config as an element/value.
+    #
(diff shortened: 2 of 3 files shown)


martinez-hub:
The failing `tests (3.9)` and `test-minimum-dependencies` checks here are unrelated to this change. They're the pre-existing, order-dependent flaky failure tracked in #574 — `MissingConfigException: In 'hydra/config': Could not find 'hydra/sweeper/basic'` in `tests/test_launch/test_validation.py` — which #876 fixes at its root cause.

All other checks pass, including `coverage`, `run-pyright`, and `tests` on 3.10–3.14. Once #876 is merged, this branch can be rebased onto `main` and CI will go fully green.

martinez-hub:
Rebased onto `main` now that #876 has merged (commit `11ade06`). The branch now includes the global-state fixture fix, so the previously-red `tests (3.9)` and `test-minimum-dependencies` checks — which were the pre-existing #574 flake — should now pass. CI re-running on `8676620`.

PR #885, opened 2026-09-18
Fix multirun on hydra-core 1.3.7: claim the sweeper target exemption

martinez-hub:
## The problem

hydra-core **1.3.7** added a target policy that refuses to instantiate any `hydra._internal.*` target named by declarative configuration. Hydra exempts its own plugins by instantiating them inside `_trusted_internal_target` (`hydra/core/plugins.py:114`), but `launch` deliberately bypasses plugin discovery — *"Instantiate sweeper without using Hydra's Plugin discovery (Zen!)"* — and calls `instantiate(cfg.hydra.sweeper)` directly. The stock `hydra/sweeper=basic` target is therefore rejected:

```
hydra.errors.InstantiationException: Reference
'hydra._internal.core_plugins.basic_sweeper.BasicSweeper' cannot be selected by
declarative configuration because it exposes implementation state. Access it
from trusted Python code instead.
full_key: hydra.sweeper
  .../hydra/_internal/target_policy.py:554
```

**Every `launch(..., multirun=True)` fails.** On hydra-core 1.3.7, `tests/test_launch` goes **28 failed / 35 passed / 4 skipped**.

## The fix

Since `launch` bypasses discovery, it claims the same exemption Hydra grants itself — scoped to the single target the config names, released immediately afterwards. It is gated behind a `HYDRA_ENFORCES_TARGET_POLICY` flag in `_compatibility.py`, following the existing `HYDRA_SUPPORTS_OBJECT_CONVERT` pattern, so **nothing changes on hydra-core < 1.3.7** (the helper is a plain `nullcontext` there).

## Verification

`tests/test_launch`:

| | hydra-core 1.3.7 | hydra-core 1.3.6 |
|---|---|---|
| before | 28 failed, 35 passed, 4 skipped | — |
| after | **63 passed**, 4 skipped | **63 passed**, 4 skipped (identical — guard inert) |

Full suite, pristine vs patched on 1.3.7, comparing failure sets: the patch **removes** the 9 launch failures and **introduces none**.

## Pre-existing failures you may want to know about

Independent of this PR, the suite has failures on hydra-core 1.3.7 that are present on a pristine checkout too. Three look like the same 1.3.6+ instantiate hardening (a callable target may no longer return a `functools.partial` subclass, which `zen_partial=True` + `zen_wrappers=` produces):

- `test_compatibility/test_hydra_supports_partial.py::test_partial_func_attr_is_always_target`
- `test_zen_processing/test_inheritance.py::test_partial_with_inherited_zen_processing`
- `test_zen_processing/test_zen_wrappers.py::test_zen_wrappers_expected_behavior`
- `test_validation.py::test_fuzz_build_validation_against_a_bunch_of_common_objects`

(`test_zen.py::test_async_compatible` and `::test_instantiation_wrapper` also fail, but that is my environment lacking an async plugin.)

Two further tests differed between runs in *both* directions; with a wiped `.hypothesis` database and a fixed seed they pass 3/3 on both pristine and patched trees, so that was example-database state, not this change.

## Alternative considered

Routing the sweeper through `Plugins.instance().instantiate_sweeper(...)` would avoid the private import entirely and work on every Hydra version — but it discards the deliberate no-discovery behaviour this line exists to provide, so I kept the current design. Happy to switch if you would rather take that trade.

The import is private (`hydra._internal.target_policy`), which is a real cost; `_launch.py` already imports `hydra._internal.callbacks`, so it is not a new kind of coupling. Version-gated either way.

Found while fixing the same breakage downstream in [mushin](https://github.com/martinez-hub/mushin), which applies the identical exemption around its own `launch` call as a stopgap and will drop it once this lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


martinez-hub:
Heads-up on this PR's red CI: **the remaining failures are pre-existing on `main`, not from this change.** The one failure that *was* mine — `check-repo-format`, a stray blank line — is fixed in `88f189e` and that job is now green.

The clearest evidence is a side-by-side with #884, a Dependabot pyright bump that only touches `/deps` and cannot involve `_launch.py`:

| job | #885 (this PR) | #884 (unrelated) |
|---|---|---|
| tests (3.9 – 3.14), test-on-windows, test-third-party, smoke-test-pydantic-v1, coverage, test-against-pre-releases | fail | fail |
| check-repo-format, run-mypy, run-pyright, pyright-scan-docs, test-docs-builds, test-minimum-dependencies | pass | pass |

Identical outcomes. `tests (3.12)` here reports `7 failed, 1600 passed` — the same pre-existing set, not the 28 `test_launch` failures this PR removes.

The timeline points at hydra-core 1.3.6:

| PR | opened | tests (3.13) |
|---|---|---|
| #880 | 2026-07-20 | pass |
| #881 | 2026-08-03 | pass |
| #882, #883 | 2026-08-31 | **fail** |
| #884 | 2026-09-14 | **fail** |

hydra-core 1.3.6 landed **2026-08-29**, between the last green PR and the first red ones, and #882/#883 are docs-only dependency bumps. 1.3.6 hardened `instantiate` so a callable target may no longer return a `functools.partial` subclass — which is what `zen_partial=True` + `zen_wrappers=` produces. Locally on hydra-core 1.3.7 the failures are:

```
tests/test_compatibility/test_hydra_supports_partial.py::test_partial_func_attr_is_always_target
tests/test_zen_processing/test_inheritance.py::test_partial_with_inherited_zen_processing
tests/test_zen_processing/test_zen_wrappers.py::test_zen_wrappers_expected_behavior
tests/test_validation.py::test_fuzz_build_validation_against_a_bunch_of_common_objects
```

Those look like a separate issue worth its own fix, and I'm happy to open one if it would help — this PR deliberately leaves them alone and only addresses the sweeper-target rejection.

For this PR specifically, `tests/test_launch` goes from **28 failed / 35 passed** to **63 passed** on hydra-core 1.3.7, and stays at **63 passed** on 1.3.6 where the guard is inert. A pristine-vs-patched diff of the full-suite failure sets shows the patch removes the launch failures and introduces none.

martinez-hub:
Filed #886 with the full diagnosis of the pre-existing CI failures referenced above — two independent causes (the `partial_with_wrapper` subclass being rejected by hydra-core >= 1.3.6's instantiate policy, and `inspect.signature` now being blocklisted as a fuzz target), with reproductions and the timeline.

Keeping them out of this PR deliberately: the natural fix for the first one removes a public class that three tests assert on, which seemed like a decision for you rather than something to attach to an unrelated bug fix. Happy to send a PR for whichever direction you prefer.

omry:
Thanks for tracking this down, and I’m sorry Hydra 1.3.7 broke `launch(..., multirun=True)`.

Hydra restricts configuration from selecting `hydra._internal` targets because those paths expose implementation objects and can be used to circumvent the security hardening. Its plugin path grants an exception after checking that the selected class is registered. Zen currently [instantiates the composed sweeper directly](https://github.com/mit-ll-responsible-ai/hydra-zen/blob/11ade066390adee051357141b30c1bead89b643f/src/hydra_zen/_launch.py#L462); the stock basic sweeper config selects an internal Hydra class. This patch instead passes the configured `_target_` to a private exemption before instantiating it. I’d prefer not to add new dependencies on Hydra’s private internals: the helper lives in `hydra._internal.target_policy` in 1.3.7 but in `hydra._internal.execution_policy` on the current 1.4 development branch.

I recommend `Plugins.instance().instantiate_sweeper(hydra_context=..., task_function=..., config=cfg)` as Hydra’s supported path for constructing a configured sweeper directly from `launch`. It checks the plugin registry, constructs the sweeper, and calls `setup()`, so Zen would not call `setup()` again. `launch` already initializes Hydra’s plugin registry during `initialize()`, so this does not add a discovery pass. In 1.3.x, custom sweepers using this path must be registered and defined under `hydra_plugins`; Hydra 1.4 adds entry-point discovery and supports explicit registration of local classes. I recognize that requiring registration would change Zen’s current support for unregistered local sweepers. If that support is essential, I’d like to work out a supported path for it rather than have Zen use the private exemption.

On #886, Hydra intentionally wraps an exact `functools.partial` returned by a target so it can check the deferred call when it is invoked, including arguments supplied at that point. It rejects partial subclasses because an overridden `__call__` can invoke behavior different from what `.func` exposes. I suggest Zen return an exact partial of a helper for the wrapped case and relax its test requiring `.func` to be the original target: Hydra’s wrapper would retain the helper as `.func`. That is a change to Zen’s tested behavior, so it should be made explicitly and covered by tests of wrapper application and the resulting call.

jgbos:
> I recommend Plugins.instance().instantiate_sweeper(hydra_context=..., task_function=..., config=cfg) as Hydra’s supported path for constructing a configured sweeper directly from launch. It checks the plugin registry, constructs the sweeper, and calls setup(), so Zen would not call setup() again. launch already initializes Hydra’s plugin registry during initialize(), so this does not add a discovery pass. In 1.3.x, custom sweepers using this path must be registered and defined under hydra_plugins; Hydra 1.4 adds entry-point discovery and supports explicit registration of local classes. I recognize that requiring registration would change Zen’s current support for unregistered local sweepers. If that support is essential, I’d like to work out a supported path for it rather than have Zen use the private exemption.

I'd recommend conforming to Hydra's plugin approach using @omry suggestion.  It does mean you have to conform to hydra_plugins, but `launch` will be more robust to future Hydra changes.

